PR #9, opened 2021-02-03
Set up for expansion of Pauli Rotations to Lattice Surgery patch operations

gwwatkin:


circuit.py
```diff
@@ -182,4 +187,25 @@ def load_from_pyzx(circuit) -> 'Circuit':
         print("Conversion completed")
         print("Gate Missed: ", gate_missed)
         return ret_circ
+
+
+    def render_ascii(self) -> str:
```
gwwatkin, line 192: This function is the reason for this PR. I was in the middle of working on something else, but I thought you'd find this function useful, so I thought I'd commit what I have.

PR #19, opened 2021-02-07
implement commute_rotation() and supporting functions

alexnguyenn:
As discussed in previous meeting.

### DIFF
--- a/rotation.py
+++ b/rotation.py
@@ -22,6 +22,39 @@ class PauliOperator(Enum):
     def __repr__(self) -> str:
         return str(self)
 
+    @staticmethod
+    def is_commute(a: 'PauliOperator', b: 'PauliOperator') -> bool:
+        """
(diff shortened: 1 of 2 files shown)


rotation.py
```diff
@@ -22,6 +22,39 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return str(self)
 
+    @staticmethod
+    def is_commute(a: 'PauliOperator', b: 'PauliOperator') -> bool:
```
gwwatkin, line 26: I was thinking we might refactor `is_commute` to be called something like `get_commute_mode`.
Maybe the return type could be an enum like
```
class CommuteMode(Enum):
    Commute = 1
    Anticommute = -1
```

gwwatkin: I get the point though ... maybe a refactoring for some other time I'd say

Issue #28, opened 2021-02-26
Complex i handling in pauli rotation commuting


gwwatkin:
In applying the transformation:
![Screenshot_20210226_092442](https://user-images.githubusercontent.com/36427091/109333571-77d41980-7814-11eb-8cc8-ab6bcf3e04dd.png)
i.e.
![Screenshot_20210226_092506](https://user-images.githubusercontent.com/36427091/109333621-85899f00-7814-11eb-9b07-4d80c308661d.png)

when evaluating the tensor product of Pauli operators `PP'`:

![Screenshot_20210226_093047](https://user-images.githubusercontent.com/36427091/109334253-5889bc00-7815-11eb-80da-6f18c85aa58e.png)

We want to replace each `P_k P'_k` with another Pauli operator, which may be multiplied by an imaginary unit (eg. `XZ=Yi`). 

**Problem:** When `P_k` and `P'_k` are distinct and not identity, imaginary units are being discarded.

**Idea:**  in:

https://github.com/alexnguyenn/lattice-surgery-compiler/blob/b27de96165d71f7a76e636c74968ddfacca0dbf7/rotation.py#L36

Change the return to `-> Tuple[PauliOperator, complex]`

The complex number is a coefficient. For this function it will always be either `1` or `complex("j")` (`i` for common mortals).

https://github.com/alexnguyenn/lattice-surgery-compiler/blob/b27de96165d71f7a76e636c74968ddfacca0dbf7/circuit.py#L119

Multiply the complex coefficients together (including the one in the formula), giving a single `c` in front of a tensor product of pauli operators to go in the exponent.  `c` has to be one of `1,-1,complex("j"),-complex("j")`

For `c=1,-1` nothing happens or the angle is adjusted. The challenge is with `complex("j"),-complex("j")`

gwwatkin:
![FF60238F-9AE1-4393-8312-0DF327E4EAF0](https://user-images.githubusercontent.com/36427091/109297770-d4dec800-7832-11eb-8ead-406cde8b7e6e.png)

gwwatkin:
Conjecture `c=complex("j"),-complex("j")` never happens when.

gwwatkin:
![B38241AF-300A-46E5-9480-05A46F45A88C](https://user-images.githubusercontent.com/36427091/109298535-099f4f00-7834-11eb-8694-c083328b1ec6.png)

gwwatkin:
[...]

PR #29, opened 2021-02-27
Addressed issue #28

alexnguyenn:
Addressing issue #28. 

Modified [PauliOperator.are_commuting()](https://github.com/alexnguyenn/lattice-surgery-compiler/commit/78f12fd5672afcb6f709b19aa97526852c72e09f#diff-ded9f751117da51f30b1392ec2d584cde2db171d132e9c51f14fb30626dbd38cR27) so that we could track number of i and adjust the phase of the rotation when needed.

### DIFF
--- a/circuit.py
+++ b/circuit.py
@@ -116,11 +116,23 @@ class Circuit(object):
 
         next_block = index + 1
         if not self.are_commuting(index, next_block):
-            for i in range(self.qubit_num):
-                new_op = PauliOperator.multiply_by_i(self.ops[index].get_op(i), self.ops[next_block].get_op(i))
-                self.ops[next_block].change_single_op(i, new_op)
+            # for tracking # of i (see issue #28)
+            # i is required for an anti-commute operators pair
+            i_count = 0     
             
+            for i in range(self.qubit_num):
+                new_op = PauliOperator.are_commuting(self.ops[index].get_op(i), self.ops[next_block].get_op(i))
+                
+                self.ops[next_block].change_single_op(i, new_op[1])
+                if not new_op[0]: i_count += 1
+            
+            # Flip the phase if the current rotation's phase is negative 
             if self.ops[index].rotation_amount < 0:
+                self.ops[next_block].rotation_amount *= -1
+
+            # Adjust the rotation based on i_count (see issue 28).
+            # It flips the phase when there is an odd number of i pairs, excluding the first i.
+            if ((i_count - 1) / 2) % 2 != 0: 
                 self.ops[next_block].rotation_amount *= -1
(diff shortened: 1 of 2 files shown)


rotation.py
```diff
@@ -22,38 +22,31 @@ def __str__(self) -> str:
...
-        return True if PauliOperator.I in (a,b) or a == b else False
 
     @staticmethod
-    def multiply_by_i(a: 'PauliOperator', b: 'PauliOperator') -> 'PauliOperator':
+    def are_commuting(a: 'PauliOperator', b: 'PauliOperator') -> Tuple[bool, 'PauliOperator']:
         """
-        Return iAB.
+        Check if 2 PauliOperators A and Bcommute or anti-commute. 
+        Returns True and AB if commute or False and iAB if anti-commute.
+
```
gwwatkin, line 31: Might be worth saying that the second returned value is C=iAB when they anticommute and C=AB when they commute

circuit.py
```diff
@@ -116,12 +116,24 @@ def commute_rotation(self, index: int) -> None:
...
             for i in range(self.qubit_num):
-                new_op = PauliOperator.multiply_by_i(self.ops[index].get_op(i), self.ops[next_block].get_op(i))
-                self.ops[next_block].change_single_op(i, new_op)
+                new_op = PauliOperator.are_commuting(self.ops[index].get_op(i), self.ops[next_block].get_op(i))
+                
+                self.ops[next_block].change_single_op(i, new_op[1])
+                if not new_op[0]: i_count += 1
             
+            # Flip the phase if the current rotation's phase is negative 
             if self.ops[index].rotation_amount < 0:
```
gwwatkin, line 130: Sorry I can't remember why we have to do this flip

gwwatkin: I suspect this was a temporary hack of some kind

gwwatkin (approved):
Just a couple notes, looks good otherwise

circuit.py
gwwatkin: 00336F56-A85A-4B27-98D3-1FB741B3EA35
When the operations commute there is no problem with the negative angle, I guess all we have to do is check the non commuting case

alexnguyenn: We don't have to worry about the commute case since we are not required to calculate iPP' (as it is just a straight swap). The line is run only when P and P' are anti-commute.

rotation.py
alexnguyenn: will update the description later

circuit.py
alexnguyenn: Further discussion in #30.

Issue #33, opened 2021-03-07
Commutation for Pauli operators pair not working as intended


alexnguyenn:
https://github.com/alexnguyenn/lattice-surgery-compiler/blob/8ecb9eb123375048f48c05a926fbf2d58bbb8826/rotation.py#L27

The function fails to take into account of the actual positions of P and P' when returning iPP'.

For example: Y = iZX != iXZ (= -Y) but PauliOperator.are_commuting() will return (False, PauliOperator.Y) for both cases.

PR #34, opened 2021-03-09
Addressing #33. Proper handling of coefficients. 

alexnguyenn:
Changes to the implementation of Pauli rotation commutation (to address #33) as follow:
- Added `PauliOperator.multiply_operators()` which will return the nearest result of PP' and the coefficients required (either `1` for commuting cases or `±i` for anti-commute cases) and  for such result. `Circuit.commute_rotation()` will adjust the phase based on the product of these coefficients. 
- Updated `Circuit.commute_rotation()` to handle commutation of an Pauli rotation block past a measurement block. Phase of rotations do not matter. Some checks in #31 has been removed based on this.
- `PauliOperator.are_commuting()` now only returns whether the 2 specified operators are commuting or not (bool). 
- Functions names and descriptions modified to reflect this change.

### DIFF
--- a/circuit.py
+++ b/circuit.py
@@ -104,46 +104,43 @@ class Circuit(object):
                 if isinstance(self.ops[index + 1], Measurement):
                     break
                 
-                self.commute_pi_over_four_rotation(index)
+                self.commute_rotation(index)
                 index += 1
 
 
-    def commute_pi_over_four_rotation(self, index: int) -> None:
+    def commute_rotation(self, index: int) -> None:
         """
-        Commute a pi/4 rotation block pass its' next block.
+        Commute a rotation block pass its neighbor block.
+
+        Args:
+            index (int): Index of the targeted block in the current circuit. 
+
         """
         next_block = index + 1
 
         if next_block >= len(self.ops):
             raise Exception("No operation to commute past")
 
-        if not isinstance(self.ops[index], Rotation) or not isinstance(self.ops[index+1], Rotation):
-            raise Exception("Can only commute rotations")
-
-        if not cast(Rotation,self.ops[index]).rotation_amount in {Fraction(1,4),Fraction(-1,4)}:
-            raise Exception("First operand must be +-pi/4 Pauli rotation")
-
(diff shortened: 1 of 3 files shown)


circuit.py
```diff
@@ -104,46 +104,43 @@ def apply_transformation(self) -> None:
                 if isinstance(self.ops[index + 1], Measurement):
                     break
                 
-                self.commute_pi_over_four_rotation(index)
+                self.commute_rotation(index)
                 index += 1
 
 
-    def commute_pi_over_four_rotation(self, index: int) -> None:
+    def commute_rotation(self, index: int) -> None:
```
gwwatkin, line 111: I think we might want to leave the pi_over_four in the name because it only applies to pi_over_four rotations

circuit.py
```diff
@@ -104,46 +104,43 @@ def apply_transformation(self) -> None:
...
 
         if next_block >= len(self.ops):
             raise Exception("No operation to commute past")
 
-        if not isinstance(self.ops[index], Rotation) or not isinstance(self.ops[index+1], Rotation):
-            raise Exception("Can only commute rotations")
-
-        if not cast(Rotation,self.ops[index]).rotation_amount in {Fraction(1,4),Fraction(-1,4)}:
-            raise Exception("First operand must be +-pi/4 Pauli rotation")
-
```
gwwatkin, line 126: I would probably keep these two checks for robustness and self documentation, I know the caller never calls them on non pi/4 rotations but it's the kind of thing helps you catch bugs

rotation.py
```diff
@@ -24,29 +26,53 @@ def __repr__(self) -> str:
...
-        if b == PauliOperator.I:
-            return True, a
+        else:
+            return True
 
-        if {a,b} == {PauliOperator.X, PauliOperator.Z}:
-            return False, PauliOperator.Y
+    
+    @staticmethod
+    def multiply_operators(a: 'PauliOperator', b: 'PauliOperator'):
```
gwwatkin, line 45: Good idea to separate the two

gwwatkin (approved):
Looks great! Just two small suggestions for robustness

Issue #35, opened 2021-03-09
Remove Y operators from pauli rotations 


gwwatkin:
We are using the compact block for logical qubits (https://arxiv.org/pdf/1808.02892.pdf sec 2.1) which doesn't support Y operators:

![image](https://user-images.githubusercontent.com/36427091/110456550-8d500b80-807e-11eb-9989-51276e316ec5.png)

gwwatkin:
The "best case" is when we have an odd number of Pauli operators in the exponent say  `i*pi/8*XYYIY`, then `XYYIY` and `IZZIZ` anticommute and as shown below we can use an identity to make to turn `i*pi/8*XYYIY` into `i*pi/8*i*(XYYIY)*(IZZIZ)` which turns into  `i*pi/8*(XXXIX)`.

In the even case we can't find an anticommuting operator to make the y's disappear, so we select one of the Y's and make just that disappear with the appropriate X, like 'YYXIYY' with 'XIIII', which leaves us with 'IYXIYY, to wich we can apply the odd case.

![Scratch_Work-14](https://user-images.githubusercontent.com/36427091/110620768-47627880-814e-11eb-9eac-73655db68a6d.jpg)

**Disclaimers** I might have lost minus signs but that doesn't affect the idea, in the code we need to make sure the signs are all correct.

PR #43, opened 2021-03-11
Implement the removeval of Y operators, circuit.remove_y_operators_from_measurement()

alexnguyenn:
Implemented removing Y operators from the circuit to address #35. Also update `Circuit.apply_transformation()` to remove pi/4 rotations once they are commuted to the end of the circuit (as an indirect implementation of pi/4 rotations absorption by measurements).

### DIFF
--- a/circuit.py
+++ b/circuit.py
@@ -104,11 +104,11 @@ class Circuit(object):
                 # if isinstance(self.ops[index + 1], Measurement):
                 #     break
                 
-                self.commute_rotation(index)
+                self.commute_pi_over_four_rotation(index)
                 index += 1
 
 
-    def commute_rotation(self, index: int) -> None:
+    def commute_pi_over_four_rotation(self, index: int) -> None:
         """
         Commute a rotation block pass its neighbor block.
 
@@ -120,6 +120,9 @@ class Circuit(object):
 
         if next_block >= len(self.ops):
             raise Exception("No operation to commute past")
+
+        if not cast(Rotation,self.ops[index]).rotation_amount in {Fraction(1,4),Fraction(-1,4)}:
+            raise Exception("First operand must be +-pi/4 Pauli rotation")
 
         # Need to calculate iPP' when PP' = -P'P (anti-commute)
         if not self.are_commuting(index, next_block):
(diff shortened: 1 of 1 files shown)


circuit.py
```diff
@@ -97,18 +93,71 @@ def apply_transformation(self) -> None:
...
+
+                    # Commute the right (new) pi/4 rotations towards the end of the circuit
+                    for right_block_index in right_block_indices:
+                        while right_block_index + 1 < len(self.ops):
+                            self.commute_pi_over_four_rotation(right_block_index)
+                            right_block_index += 1
+                        self.ops.pop()
+            else:
+                # This is assuming pi/4 rotations (not including ones from this operation) 
+                # have been commuted to the end of the circuit.
```
gwwatkin, line 154: Might be my shortcoming, but it took me a while to figure out that there was another assumption: that the Litinski transform has already been applied.

circuit.py
```diff
@@ -97,18 +93,71 @@ def apply_transformation(self) -> None:
...
+                break
```
gwwatkin, line 155: Now that I think about it more carefully, would it be very challenging to make this transformation work without assuming that the Litinski Transform has already been applied? There might be reasons why we wouldn't want to always apply it, like if we are debugging other stuff and we still want operations to stay.

gwwatkin: Intuitively I would think that all we have to do is turn the break into a continue and increase the counter
```suggestion
                continue
```

gwwatkin: Of course this doesn't work for pi/4 rotations, but I guess that's ok for now we can deal with thatas it becomes necessary, since it's as simple of a change

gwwatkin (approved):
Looks good, a small suggestion. Got to remember #44: to check that the way measurement is handled is conceptually correct

circuit.py
alexnguyenn: Originally `apply_transformation()` does not remove the `pi/4` rotations after they are moved to the end of the circuit, that is why I left the the `break` comment there just to terminate the while loop early. Implement removing the `pi/4` rotations was the last minute decision - I was conflicted on whether or not to remove information from the circuit (in case we want to reverse the operation).

alexnguyenn: Anyway it was me overthinking stuffs - discussions adding options to reverse `apply_transformation()` can happen later. I will go ahead and merge the branch.

PR #46, opened 2021-03-13
Simulate underlying patch states

### LINKED ISSUE #40: Simulation of patch states

gwwatkin:
Can track the state of the logical computation in of `qiskit.aqua.operators.DictStateFn`. So the computation is represented as one big quantum register that evolves with the slices. That is for each slice there is one `DictStateFn` object representing the state of the system.

The simulation happens in parallel with state generation, which makes conditional corrective not to appear on the lattice when they are not required by the previous measurement outcomes. This is the only visible impact of simulation on the UI.

closes #40

### DIFF
--- a/conditional_operation_control.py
+++ b/conditional_operation_control.py
@@ -16,16 +16,8 @@ class HasPauliEigenvalueOutcome():
 
 
 class EvaluationCondition:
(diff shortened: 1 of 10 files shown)


gwwatkin:
Changes to the UI split to #47.

alexnguyenn (approved):

alexnguyenn (approved):

PR #63, opened 2021-03-24
Web style

Keelando:
First major change to website gui style. Nothing is permanent, and I am highly open to feedback.

PR #64, opened 2021-03-24
Styling + Deployment fixes followup

gwwatkin:
Follows up with small changes #63


.gitignore
```diff
@@ -5,3 +5,5 @@ webgui/deploy.log
 webgui/env
 webgui/server_pid
 index.html
+Pipfile
+Pipfile.lock
```
gwwatkin, line 9: Removed these files because they always change and pollute diffs. It doesn't appear that they are used in deployment

gwwatkin:
Also added some fixes for deployment

web/lattice_surgery_web/settings.py
```diff
@@ -79,12 +79,12 @@
...
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': BASE_DIR / 'db.sqlite3',
-    }
-}
+# DATABASES = {
+#     'default': {
+#         'ENGINE': 'django.db.backends.sqlite3',
+#         'NAME': BASE_DIR / 'db.sqlite3',
+#     }
+# }
```
gwwatkin, line 87: Gettin the error:
```
[...]
  File "/var/www/vhosts/lattice-surgery-compiler/webgui/env/lib64/python3.8/site-packages/django/db/utils.py", line 111, in load_backend
    return import_module('%s.base' % backend_name)
  File "/usr/lib64/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/var/www/vhosts/lattice-surgery-compiler/webgui/env/lib64/python3.8/site-packages/django/db/backends/sqlite3/base.py", line 70, in <module>
    check_sqlite_version()
  File "/var/www/vhosts/lattice-surgery-compiler/webgui/env/lib64/python3.8/site-packages/django/db/backends/sqlite3/base.py", line 67, in check_sqlite_version
    raise ImproperlyConfigured('SQLite 3.8.3 or later is required (found %s).' % Database.sqlite_version)
django.core.exceptions.ImproperlyConfigured: SQLite 3.8.3 or later is required (found 3.7.17).
```

gwwatkin: @Keelando

gwwatkin: We could use mysql in the future

gwwatkin: Or there is this solution https://stackoverflow.com/questions/56520658/djangoshowing-improperlyconfigured-error-sqlite-3-8-3-or-later-is-required-f

gwwatkin: Spent a lot of time on this. Tried to run a custom version of sqlite, but in some cases the versions still don't match. Will have to either:
- Build python from scratch
- Downgrade django
- Containerize the application

It is a well known bug

gwwatkin: Or fourth option one can insist with a custom compiled sqlite https://stackoverflow.com/questions/55674176/django-cant-find-new-sqlite-version-sqlite-3-8-3-or-later-is-required-found

.gitignore
Keelando: This may be moot if we dockerize, but pipenv can install from either requirements.txt or Pipfile. Requirements.txt needs to be periodically manually updated with a pipenv lock -r > requirements.txt, whereas the Pipfile reflects which modules have been currently added by pipenv. As long as the requirements.txt include locked version numbers, we can probably remove Pipfile and Pipfile.lock from version control.

web/lattice_surgery_web/settings.py
Keelando: Hmm. I'm running sqlite3 version 2.6.0. Try pip install sqlite3?

.gitignore
gwwatkin: Ok I see now, I thought those files were just a byproduct.`requirements.txt` is easier to edit for humans so might be worth sticking with that. Usually I would have seen this sort file handled manually. I think for us the requirements file could be as short as:
```
-i https://pypi.org/simple

django==3.1.7
django-crispy-forms==1.11.2
pyzx==0.6.4
python-igraph==0.9.0
qiskit==0.24.0; python_version > 3.6

```
And let dependency resolution sort out the rest on each system, locking other packages as need arises.

But I'm not really sure about what the practices are with python projects.

web/lattice_surgery_web/settings.py
gwwatkin:
```
ERROR: Could not find a version that satisfies the requirement sqlite3>=2.6.0 (from -r requirements.txt (line 127)) (from versions: none)
ERROR: No matching distribution found for sqlite3>=2.6.0 (from -r requirements.txt (line 127))
```
It seems that sqlite3 is not a package for python 3.8

.gitignore
Keelando: If you are unable to get pipenv on your server machine, the Pipfiles won't be very helpful I guess. It's just a statement of which sub-dependencies are used, for example, in qiskit, you wouldn't believe how many dependencies that one has, and the pipfiles show which ones were installed. When we choose another library to add to the project, pipenv will see which sub-dependencies are being used and if there is a conflict, it will try other possible combinations. If you install the packages one by one, you might end up with incompatible versions. It's not the end of the world if we can't use pipenv, ill add Pipfile and Pipfile.lock to gitignore if you haven't already. Locked versions and consistent ordering of libraries should be okay for now.

Keelando (approved):

Keelando:
That's funny, I never thought of using {## } for comments in the templates, but it seems to work fine. Django python blocks in the template are usually {{ }} to render context variables or {% %} to render for code such as loops etc. {%#  %} is another way to comment out stuff.

gwwatkin:
PyCharm puts these comments in automatically when using the comment shortcut

.gitignore
gwwatkin: Ok I propose a hybrid approach: We remove the pip files from the gitignore, but handle manually the requirements.txt, this way we have both human readable depndency and, as needed - even if not used in deployment,a complete snapshot in the Pipfiles.

gwwatkin: We just leave the rule that Pipfile changes allways deserve their own commit

gwwatkin: To keep the diffs nice

PR #65, opened 2021-03-26
Small layout changes

Keelando:
Going to leave this one open as I make a few more changes before switching gears to other features. I will have a look at rendering the output grid in a way that maximizes its readability and avoids formatting errors. I might be able to make it resizable.


web/lattice_main/templates/lattice_main/base.html
```diff
@@ -28,60 +28,60 @@
 
 <body>
   <div style="background-color: black; padding:10px">
-      <h1 class='aquire' style="color:#1dc700;text-align: center;">Lattice Surgery Quantum Error Correction Compiler</h1>
+      <h1 class='aquire' style="color:#E0E0E0;text-align: center;">Lattice Surgery Quantum Error Correction Compiler</h1>
```
gwwatkin, line 31: ahahah no more terminator

web/lattice_main/render_to_file.py
```diff
@@ -0,0 +1,31 @@
+#### Stand-alone script version of render. (not yet sure why the CSS won't render)
```
Keelando, line 1: I wrote this as a stand-alone script, however, I'm not sure what output you are expecting, and the CSS doesn't render. I also integrated it into a view that is callable by URL. Maybe we want to integrate saving html outputs into the GUI? Looking for guidance on how you were using this in your debugging.

gwwatkin: Thank you keelan this is very helpful I think the reason why the css won't render is because it's served on another domain

gwwatkin: Maybe this helps https://stackoverflow.com/questions/22162027/how-do-i-generate-a-static-html-file-from-a-django-template 

Anyways the reason why we care is for debugging, so we can run the `debug_*` files directly from the ide with the debugger, without having to start the server with the debugger enabled

gwwatkin: So no need for integration of this tool in the gui, it's just for debugging

Keelando: Ok, I'll remove it from the Views then

Keelando:
Just some small changes. Still haven't touched the output circuit styling which will come on the next PR

web/lattice_main/render_to_file.py
```diff
@@ -0,0 +1,36 @@
...
+    rendered_page = render_to_string("lattice_main/lattice_view.html",context)
+    with open(output_file_name, 'w', encoding="UTF-8") as f:
+        f.write(rendered_page)
+
+DEMO_CIRCUITS_DIR = "../../assets/demo_circuits/"
+
+circuit = "nontrivial_state.qasm"
+output_filename = "saved.html"
+
+render_to_file(DEMO_CIRCUITS_DIR + circuit, output_filename)
```
gwwatkin, line 36:
```suggestion
if __name__ == "__main__":
    render_to_file(DEMO_CIRCUITS_DIR + circuit, output_filename)
```
To prevent the command from being executed when the file is just imported

web/lattice_surgery_web/settings_template_test.py
```diff
@@ -0,0 +1,126 @@
+"""
```
gwwatkin, line 1: Maybe this file was committed accidentally?

web/lattice_main/views.py
```diff
@@ -51,20 +52,28 @@ def view_compiled(request):
     }
     return render(request,"lattice_main/lattice_view.html",context)
 
-# TODO restore this for debugging
-# def render_to_file(request,slices, output_file_name): # (slices : List[patches.Lattice], output_file_name : str) -> str
-#     slices = list(map(sparse_lattice_to_array, slices))
-#     styles_map = styles_map
-#     patches = patches
+# def render_to_file(request) -> str:
```
gwwatkin, line 55: I suppose this commented out code could go now?

gwwatkin (approved):
Looks good, a couple minor suggestions

web/lattice_surgery_web/settings_template_test.py
Keelando: This is actually an alternate settings.py file which is used when render_to_file.py is called. In render_to_file, the django environment is loaded but I needed to to specify full paths for the static files so that it finds the static files such as css and images on your local browser when you debug. Perhaps it should be renamed according to how you use the render to file script as a debugger. Maybe settings_render_to_file.py?

web/lattice_main/views.py
Keelando: Indeed. I suppose I can always dig it up mining old commits

web/lattice_surgery_web/settings_template_test.py
Keelando: changed it to settings_render_to_file.py

Issue #50, opened 2021-03-16
Add support for simple conditional operations of QASM


gwwatkin:
Support statements like:
```
measure q[0] -> c0[0];
measure q[1] -> c1[0];
if (c0==1) x q[2];
```
Useful for things like the teleportation protocol.

Examples like the one above can already be implemented at the circuit level using [EvaluationCondition](https://github.com/latticesurgery-com/lattice-surgery-compiler/blob/974073938483b3535046deb16552b6c259d4b368/conditional_operation_control.py#L16) and related classes.

This issue is mainly related to parsing. A technique similar to that of #49, using qiskit's  AST for QASM, should do.

Lower priority.

Issue #49, opened 2021-03-16
Add support for measurement operators in QASM


gwwatkin:
A quantum circuit may have measurements before the end of the computation. Since PYZX only supports reversible computation, one must manually handle the measurement.

The plan is to pre parse the input circuit with qiskit (this may come at the cost of dropping support for non QASM circuits) and break it into reversible sections, to be handed over to pyzx for further processing, and measurements, to be read directly from qiskit's abstract syntax tree.

PR #74, opened 2021-04-07
Teleportation Example

gwwatkin:
A hard coded teleportation example


gwwatkin:
Keeping for reference but moving but restructuring to integrate #49 #50

PR #76, opened 2021-04-08
Salvage what was good of the teleportation example

gwwatkin:
Halting the development of hard coding a teleportation example in favor of making it work directly from a QASM file with  #49 and #50.

saves things from #74 .


gwwatkin:
@Keelando  for review for the change in the `settings.py` file.

alexnguyenn (approved):

PR #78, opened 2021-04-09
Contact form

Keelando:
Its time to think about the contact emailer. Did you still want to set up an email @latticesurgery.com?

Since we're not sending out newsletters, we just need the contact emails forwarded to our own email addresses. For this purpose, it doesn't matter what the email domain is. I can set up a gmail emailer to forward the message to us in an email.
Unless you want people getting email confirmation that their message went through, in which case it would be nice to have @latticesurgery.com as the origin.

I could also hard-code emails in the web page, but the contact form might be better for security of email accounts.

Another thought is to provide a snippet about each contributer, like just name and github or something, like a developer profile sort of thing.


alexnguyenn:
Closing this since it is not applicable to this repo anymore.

PR #80, opened 2021-04-12
Option to disable Y operator removal

gwwatkin:
Used to make the example in the presentation where the stabilizer removal transform was applied without removing the Y operators from Pauli rotations. Probably worth keeping.


gwwatkin:
Moved from #79

alexnguyenn (approved):

PR #87, opened 2021-04-24
Add method to render circuit in latex 

alexnguyenn:
Currently functional. Missing phase labelling and adaptive sizing (can be difficult to implement). May have difficulties with long circuit due to limited width of document. The method uses `mako` package for template library.


gwwatkin (approved):

gwwatkin:
This is what this rendering looks like:
image

PR #89, opened 2021-04-24
Add phase labelling for latex rendering 

alexnguyenn:
As title. Below is the screenshot of the latex rendering for the circuit in `debug/debug_circuit_render_latex.py`.

image

### DIFF
--- a/assets/circuit_latex_render.mak
+++ b/assets/circuit_latex_render.mak
@@ -6,7 +6,9 @@
 
 \begin{document}
 
-\begin{tikzpicture}
+\hoffset=-1in
+\voffset=-1in
+\setbox0\hbox{\begin{tikzpicture}
     \tikzstyle{paulicomponent} = [draw,draw=none,fill=white,minimum size=1.5em] 
     \tikzstyle{phase} = [draw,fill,shape=circle,minimum size=5pt,inner sep=0pt]
     \matrix[row sep=0.4cm, column sep=0.4cm] (circuit) {
@@ -46,5 +48,10 @@
         ;
     \end{pgfonlayer}
     
-\end{tikzpicture}
+\end{tikzpicture}}
+\pdfpageheight=\dimexpr\ht0+\dp0\relax
+\pdfpagewidth=\wd0
+\shipout\box0
+
+\stop
 \end{document}
--- a/assets/circuit_latex_render.mak
+++ b/assets/circuit_latex_render.mak
@@ -15,9 +15,9 @@
     
     ## Generate each row:
     % for i in range(qubit_num):
-        \node (q${i}) {$q_1$}; &[-0.1cm]
+        \node (q${i}) {$q_${i}$}; &[-0.1cm]
         % for j in range(i, len(operator_list), qubit_num):
-        \node[paulicomponent] (H${j}) {${operator_list[j]}}; &
+        \node[paulicomponent] (op${j}) {${operator_list[j]}}; &
         %endfor 
         \coordinate (end${i}); \\\
 
@@ -36,7 +36,7 @@
 
     ## Drawing borders for each Pauli operation:
     % for k in range(0, len(operator_list), qubit_num):
-    \draw (H${k}.north east) rectangle (H${(k+qubit_num-1)}.south west);
+    \draw (op${k}.north east) rectangle (op${(k+qubit_num-1)}.south west);
     % endfor 
     
     ## Drawing circuit lines: 
--- a/assets/circuit_latex_render.mak
+++ b/assets/circuit_latex_render.mak
@@ -19,15 +19,22 @@
         %endfor 
         \coordinate (end${i}); \\\
 
-        ## Won't need the below line later (when adding phase labels)
-        % if i == qubit_num - 1: 
-    }; 
-        % endif 
+    % endfor
+        [-0.3cm]
+        ##
+        ## Phase labelling:
+        ##
+        \node (op_angle_begin) {}; &
+    % for phase_label in phase_list:
+        \node[paulicomponent] {${phase_label}}; &
+    %endfor
+        \coordinate (op_angle_end); \\\
+
+    };
 
     ## Drawing borders for each Pauli operation:
-    % endfor
     % for k in range(0, len(operator_list), qubit_num):
-        \draw (H${k}.north east) rectangle (H${(k+qubit_num-1)}.south west);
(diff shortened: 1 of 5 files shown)


alexnguyenn:
Added auto page sizing based on this [thread](https://tex.stackexchange.com/questions/299005/automatic-page-size-to-fit-arbitrary-content). Note that it only resize the latex page so it can display the whole circuit, not the circuit rendering itself.

alexnguyenn:
Rendering from Litinski example circuit (which can be found in `debug/debug_litinski_transform_seminar_example.py`):
image

alexnguyenn:
978cbc4: add `PauliProductOperation.to_latex()` to allow users to export equation form (latex) of a Pauli Product operation. Might be useful in the future.

rotation.py
```diff
@@ -147,7 +156,19 @@ def __init__(self, no_of_qubit: int, rotation_amount: Fraction) -> None:
...
             
-        
+
+    def to_latex(self) -> str:
+        return_str = super().to_latex() 
+        return_str += '_\\frac{'
+        if abs(self.rotation_amount.numerator) != 1:
+            return_str += str(self.rotation_amount.numerator)
+        elif self.rotation_amount.numerator == -1:
+            return_str += '-'
```
gwwatkin, line 166: Is there a particular reason why we are adding the minus sign inside the numerator? like it messes the layout otherwise? Normally you would have the minus sign in front of the fraction.

circuit.py
```diff
@@ -330,11 +330,19 @@ def render_latex(self) -> str:
...
                 operator_list += str(operator) 
             
+            # Latex format for phase label (I didnt want to do this in the template file)
             if isinstance(operation, Rotation):
-                phase_list.append((operation.rotation_amount.numerator, operation.rotation_amount.numerator)) 
+                operator_str = '$\\frac{'
+                if abs(operation.rotation_amount.numerator) != 1:
+                    operator_str += str(operation.rotation_amount.numerator)
+                elif operation.rotation_amount.numerator == -1:
+                    operator_str += '-'
```
gwwatkin, line 339: Same comment about minus sign placement here

assets/circuit_latex_render.mak
```diff
@@ -19,15 +21,22 @@
...
+    %endfor
+        \coordinate (op_angle_end); \\\
+
+    };
 
     ## Drawing borders for each Pauli operation:
-    % endfor
     % for k in range(0, len(operator_list), qubit_num):
-        \draw (H${k}.north east) rectangle (H${(k+qubit_num-1)}.south west);
+    \draw (H${k}.north east) rectangle (H${(k+qubit_num-1)}.south west);
```
gwwatkin, line 39: The H here was arbitrary just because there we H operators in the example I took ... might be worth making the name more informative

assets/circuit_latex_render.mak
```diff
@@ -6,7 +6,9 @@
 
 \begin{document}
 
-\begin{tikzpicture}
+\hoffset=-1in
+\voffset=-1in
+\setbox0\hbox{\begin{tikzpicture}
     \tikzstyle{paulicomponent} = [draw,draw=none,fill=white,minimum size=1.5em] 
     \tikzstyle{phase} = [draw,fill,shape=circle,minimum size=5pt,inner sep=0pt]
     \matrix[row sep=0.4cm, column sep=0.4cm] (circuit) {
```
gwwatkin, line 14: Just below here at line 16=>18 
`\node (q${i}) {$q_1$}; &[-0.1cm]` I think you meant `\node (q${i}) {$q_`**${i}**`$}; &[-0.1cm]`

gwwatkin: Can't make a suggestion because it's not in the diff

gwwatkin (approved):
Looks good, just a couple of small comments.

Was thinking ... maybe we can colour the rotations and measureemnts, like with some light blue and yellow?

assets/circuit_latex_render.mak
alexnguyenn: good catch

rotation.py
alexnguyenn: if the numerator is -1/1 i want it to display as -pi/pi on the graph.

gwwatkin: I think this does what you want while keeping the minus sign out of the fraction
```
from fractions import Fraction

def render_phase_frac_to_latex(phi : Fraction):
  """Assumes phi is multiplied by pi"""
  if phi.numerator== 0: return "0"

  sign = "" if phi.numerator > 0 else "-"
  num = "" if abs(phi.numerator) == 1 else str(abs(phi.numerator))

  if phi.denominator == 1:
    return "%s%s\\pi"%(sign,num)

  den = str(phi.denominator)

  return "%s\\frac{%s\\pi, %s}"%(sign,num,den)

```
Example:
```
if __name__ == "__main__":
  print(render_phase_frac_to_latex(Fraction(0,1)))
  print(render_phase_frac_to_latex(Fraction(1,1)))
  print(render_phase_frac_to_latex(Fraction(-1,1)))
  print(render_phase_frac_to_latex(Fraction(1,2)))
  print(render_phase_frac_to_latex(Fraction(-1,2)))
  print(render_phase_frac_to_latex(Fraction(1,4)))
  print(render_phase_frac_to_latex(Fraction(-1,4)))
  print(render_phase_frac_to_latex(Fraction(1,8)))
  print(render_phase_frac_to_latex(Fraction(-1,8)))
  print(render_phase_frac_to_latex(Fraction(7,8)))
  print(render_phase_frac_to_latex(Fraction(-7,8)))
```
Outputs
```
0
\pi
-\pi
\frac{\pi, 2}
-\frac{\pi, 2}
\frac{\pi, 4}
-\frac{\pi, 4}
\frac{\pi, 8}
-\frac{\pi, 8}
\frac{7\pi, 8}
-\frac{7\pi, 8}
```

circuit.py
gwwatkin: Since there are two of these sections, might be a good idea to refactor into a function too

rotation.py
alexnguyenn: implemented in 6e4fe5f.

assets/circuit_latex_render.mak
alexnguyenn: Changed `H` to `op` in 6e4fe5f.

Issue #93, opened 2021-04-30
Add Transversal Operations


gwwatkin:
After discussion with @alexandrupaler , we convinced ourselves that what we want for a paper is comparing direct compilation and compilation with the Stabilizer Commuting Transform.

Now compilation is centered around the idea of going through the curcuit of pauli rotations, which is meant for the SCT. Without the SCT we don't need necessarily this representation, and we can perform directly (transversally) certain operations, like Hadamards and CNOTs.

So in the meeting in which we discussd #91, we decided that we were going to add support for Transversal operations at Circuit level. But this presents a problem because the Circuit is geared for pauli product operations.

**Step 1:** Rename Circuit to PauliCircuit.

Now we need a Circuit equivalent for transversal operations. 

**Step 2:** Make an equivalent to the old circuit for Transversal operations.

Some operations like T and S gates, will still require corrective terms and ancilla states. To avoid re programming those steps, T gates and S gates will be handled recycling the Circuit functionality, which already handles ancilla and  corrective terms.

**Step 3:** Implement S and T gates reusing the functionality from Circuit

_Note:_ X,Y,Z can also be implemented transversally, so it needs to be evaluated if those should reuse circuit as well or just be considered pi/2 rotations

next we need to reconnect to the next step in the pipeline:

**Step 4:** Extend logical lattice operations to include Transversal CNOT and H.

_Note:_ pi/2 rotations are already compiled down to transversal operations in: https://github.com/latticesurgery-com/lattice-surgery-compiler/blob/ae4a9dc9ebde1c8c38f20fb51b2dc60ee21c8298/logical_lattice_ops.py#L47

Issue #82, opened 2021-04-18
Parallelize operations where possible


gwwatkin:
In #71 we started working to this task. Now we want to be able to execute as many operations as possible in parallel.

This task is very important towards a potential publication, because the ability to parallelize will let us evaluate the effectiveness of different optimization passes. For example, clearing the stabilizer part of the circuit can reduce the circuit depth, but it may compromise how much a circuit can be parallelized. This because after the pass the non stabilizer operations seem to span more qubits and tend to overlap more.

The general outline of the process is the following:
-  Identify parallelizeable operations (those that commute + those that have explicit conditional relations)
- Create a DAG of operation dependencies (Done in #71)
- Compile the DAG parallelizing as much as possible. Besides dependency  another limiting factor limiting parallelization is routing of ancilla regions which can't overlap.

gwwatkin:
## Proposed pipeline
I propose this compilation pipeline to achieve parallelization. The key idea is that we replace steps happening on a linear list with those that operate on the DAG.
### Intermediate representations

  **(1)**  Circuit
  **(2)**  DAG of PauliProductOperations
  **(3)**  DAG of LogicalLatticeOperation
  **(4)**  LatticeSurgeryComputation (slices)

### Stages

**(1)**->**(2)** DAG building of `PauliProductOperation`s: Done. 

**(2)**->**(3)** Going from `PauliProductOperation`s to `LogicalLatticeOperation`s within all in DAG form. In the existing linear step for this is in `LogicalLatticeComputation._load_circuit()`. 

This compilation step introduces new terms: where there was a Pauli Rotation node, it has to be replaced by a subgraph of a multi body measurement and corrective terms depending on it, which may have their own sub corrective terms. So both the underlying operations and the structure of the dag change here.

I propose two approaches:
 - Make a new class, say `ConcurrentLogicalLatticeComputation` where instead of having `self.ops` as `list`, it's a DAG. If we go with this approach, then it's probably advantageous to go to `DependencyGraph.Node`, make `self.op` a pointer to the operation rather than the index of the operation. In stage (2) `self.op` would point to a `PauliProductOperation` and in stage (3) to a `LogicalLatticeOperation`.
[...]

PR #94, opened 2021-05-03
Use operation DAG to Split LogicalLatticeComputation into Linear and Parallel

gwwatkin:


gwwatkin:
On Hold. Opting for simpler scheduling for the time being.

Intended to satisfy #82. But it's going to need more restructuring. Given that we are going to add #93 we probably want to complete the regular workflow of transversal operations before jumping in with a DAG of those too.

Issue #96, opened 2021-05-12
Compilation: Litinski vs Non-Litinski comparison.


gwwatkin:
## Targets

**Litinski Compilation:** The one we've been following so far: goes through pauli rotations and multibody measurement with ancilla regions (for which the details of the stabilizer measurements are not 100% clear yet). Includes the stabilizer commuting transform.

**Non-Litinski**: based on the original lattice surgery paper, [Horsman et al. 2013](https://arxiv.org/abs/1111.4022).

The first kind of compilation is the one we already have. What we need to do next is implement operations as they are given by horseman et all. At least Pauli operations and CNOTs. Started add transversal ops in #91 and #93. For CNOTs, they are not transversal in Horsman et al. so we might want to find a way of expressing that too.

## Benchamrk circuits

We want to see with realistic circuits how these two compare.

Three levels of real world circuit to benchmark on (from less to more structured):

**NISQ:** algorithms like QAOA and VQE.

**Intermediate:** Quantum random walks, Quantum Neural Networks.

**Structured:** Oracles with arithmetic (eg. with Karatsuba multipliers),  Shor's.

## Simulation and correctiver terms

A the moment we are simulating states to obtain corrective terms that follow multibody measurements ([Litinski, sec 1.1](https://arxiv.org/pdf/1808.02892.pdf)). It probably is desirable to run some tests without simulation. We could extrapolate from a small scale example the frequency of these terms and then apply them randomly; alterbnatively consider the worst case and always apply them.

gwwatkin:
@alexandrupaler drawing of the comparison. We looked at the idea by trying the example of a Toffoli gate:
![image](https://user-images.githubusercontent.com/36427091/118162151-47dcf100-b3d5-11eb-94e4-1e8479359c4c.png)

gwwatkin:
Trying to sum up @alexandrupaler's discussion: The essential idea is that on the fly parallelization might be more effective in Horseman's  formalism. Optimizing Litinski's Pauli rotations becomes a combinatorial problem that could be hard to solve on the fly. On the other hand some circuits (like potentially the Toffoli gate in the example) naturally fall into a parallelizeable configuration.

Issue #91, opened 2021-04-29
Define Transversal Operations Requirements


gwwatkin:
- Research which operations can be done transversally and in how many timesteps
 - Describe how compilation is impacted:
      Are we doing this?
```
class TransversalizedCircuit:
    def __init__(...):
        self.ops : List[Union[Circuit, Tuple[int,TransversalOp]]] = []
```

gwwatkin:
An idea, recycle the pauli operation infrastructure to do S and T gates

Clifford set:
   - X, Z -> Transversal, already implemented at the next level:
       https://github.com/latticesurgery-com/lattice-surgery-compiler/blob/ae4a9dc9ebde1c8c38f20fb51b2dc60ee21c8298/logical_lattice_ops.py#L47
   - Y -> avoid it
   - CNOT -> Transversal, make it into a `LogicalCNOT(source_uuid, target_uuid)`
   - H -> Transversal, make it into a `LogicalHadamard(target_uuid)`
   - S, T -> Not trasnversal, it requires ancillas.
   
**Solution for the S and T:**
Reuse the infrastructure that complies to `LogicalLatticeOperations` going through `PauliCircuit` to reuse the ancilla handling infrastructure.

**Overview**
![Untitled Diagram](https://user-images.githubusercontent.com/36427091/116600272-fa915780-a8dd-11eb-905c-7c63f1f17eb2.png)

gwwatkin:
"Generalized circuit" or maybe TrasversalizedCircuit is what holds `self.ops : List[Union[Circuit, Tuple[int,TransversalOp]]]`

gwwatkin:
The problem here is that we are not interested in transversal CNOT. We want to do CNOT by lattice surgery, but not going through Pauli rotations. So the word Transversal is not correct.

PR #97, opened 2021-05-12
Add non Litinski Operations

gwwatkin:
This is part of the bigger task #96.

Initially we thought in #91 and #93 to add just Transversal operations, but in the Horsman et al. paper there is a precise protocol for CNOT that doesn't require converting to PauliRotations before lattice surgery. We could use that directly. That means adding another kind of operation, which is  neither transversal nor Litinski. It's the original kind of lattice surgery.

### DIFF
--- a/circuit.py
+++ b/circuit.py
@@ -4,6 +4,7 @@ from rotation import PauliProductOperation, Rotation, Measurement, PauliOperator
 from fractions import Fraction
 from utils import decompose_pi_fraction, phase_frac_to_latex
 import pyzx as zx
+from transversal_op import *
 from typing import *
 
 class PauliCircuit(object):
@@ -21,7 +22,7 @@ class PauliCircuit(object):
             name (str, optional): Circuit's name (for display). Defaults to ''.
         """
         self.qubit_num:     int = no_of_qubit
-        self.ops:           List[PauliProductOperation] = list()
+        self.ops:           List[Union[Circuit, TransversalOp]] = list()
         self.name:          str = name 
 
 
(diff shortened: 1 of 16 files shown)


circuit.py
```diff
@@ -4,24 +4,25 @@
...
-        Generating a circuit
+        Generating a Pauli circuit
 
         Args:
             no_of_qubit (int): Number of qubits in the circuit
             name (str, optional): Circuit's name (for display). Defaults to ''.
         """
         self.qubit_num:     int = no_of_qubit
-        self.ops:           List[PauliProductOperation] = list()
+        self.ops:           List[Union[Circuit, TransversalOp]] = list()
```
gwwatkin, line 25: In #93 we decided to go with:
Screenshot_20210512_070859
I thought that would have meant leaving the PauliCircuit class separate from transversal operations. So the mixing of transversal and non transversal would happen outside of the circuit class of which we would take segments :thinking: . Is there a particular reason why you went this way?

alexnguyenn:
@gwwatkin Any idea on the status of this? Do we still need this feature?

gwwatkin:
@alexnguyenn I would say it's been on hold for some time. Now I'm inclined to replace it with #116 as a more more relevant and up to date approach to direct compiling

gwwatkin:
Updated #96 indicating #116 - compiling directly ZX to Lattice surgery as higher priority. If we want to implement Horsman style operations we'll have to rewrite this anyways because of more recent changes. There will still be a trace of this PR in #96 conversation if we need inspiration, so I think we can safely close.

PR #110, opened 2021-10-09
Lattice View JSON API + Set up local development API server

### LINKED ISSUE #109: Lattice view JSON api
The lattice view page renders HTML based on the contents of a nested python array of slices made of a grid of cells:

https://github.com/latticesurgery-com/lattice-surgery-compiler/blob/aa3d24a958d5ab84424d0b482d6090a5fc51eb2d/sparse_lattice_to_array.py#L7

To decouple the compiler (#106) we want to be able to send this nested array across the network. The goal is to have something that can be easily used to recreate the render loop below, that now lives in a template, and that can easily be sent and received.
https://github.com/latticesurgery-com/lattice-surgery-compiler/blob/aa3d24a958d5ab84424d0b482d6090a5fc51eb2d/webgui/templates/lattice_view.mak#L89-L94

The format of choice for this sort of thing is usually JSON. The array would look something like:
```JSON
{ 
  "slices" : [
     [
       [ 
          { "type" : "qubit",
            "edges" : [
            ],
             "text" : "|0>"
          },
          { "type" : "blank",
            "edges" : [
            ],
             "text" : ""
          },
       ],
       [ 
          { "type" : "qubit",
            "edges" : [
            ],
             "text" : "|1>"
          },
          { "type" : "blank",
            "edges" : [
            ],
             "text" : ""
          },
       ]
     ]
  ]
}
```
But with all the details of `VisualArrayCell`
* Required by #106

gwwatkin:
closes #109 

### How to use:
Start the local server by running `$ python webapi/local_compiler_api_server.py`. This serves at `localhost:9876` and has only one path, `POST  http://localhost:9876/compile` which accepts a json and returns a json. As an example and for debugging, `local_compiler_api_server.py` [shows](https://github.com/latticesurgery-com/lattice-surgery-compiler/pull/110/commits/3845f1a9928dd363855bde9a61e84a453d39ffe5#r725521905) how to call this API from a python script using `http.client`.

However the api is designed to be called from browser, with jquery for example it could be:

```js
$.post(`htttp://localhost:9876`, {
    'circuit_source': 'str', // alternative would be 'file' but is not yet supported
    'circuit': SAMPLE_CIRCUIT,
    'apply_litinski_transform': True
}, ( function(data) {
   // Populate the HTML to form the lattice view.
} )
```

To do so, it might be necessary to set the [`Access-Control-Allow-Origin`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) HTTP fields on the server serving the lattice view page.

### Overview of the API
Input is passed as JSON parameters:
```js
const SAMPLE_CIRCUIT = """
OPENQASM 2.0;
include "qelib1.inc";
qreg q[2];
h q[0];
cx q[0],q[1];
h q[0];
t q[1];
s q[0];
x q[0];
"""

// These are the post parameters to be passed to the API. See above for how to use with JQuery or in the PR for python.
const request_data = {
    'circuit_source': 'str', // alternative would be 'file' but is not yet implemented
    'circuit': SAMPLE_CIRCUIT,
    'apply_litinski_transform': true // Optional
}
```
Output is generated automatically by traversing the array of GUISlices:
https://github.com/latticesurgery-com/lattice-surgery-compiler/blob/f0ec1c77c6bdfa0dc89ea4252433f46b03fba5c1/lattice_surgery_compilation_pipeline.py#L12) 
And looks something like: 
```js
{
  "slices": [
    [
      [
        {
          "edges": {
            "Top": "Dashed", 
            "Bottom": "Dashed", 
            "Left": "Solid", 
            "Right": "Solid"
          }, 
          "patch_type": "Qubit", 
          "text": "|0>", 
          "activity": null
        },
     ], // ... 
   ], // ...
  ], // ...
"compilation_text": "Input Circuit:\n ... "
}
```

### DIFF
--- a/lattice_surgery_compilation_pipeline.py
+++ b/lattice_surgery_compilation_pipeline.py
@@ -3,6 +3,7 @@ import segmented_qasm_parser
 import lattice_surgery_computation_composer
 import logical_lattice_ops
 import sparse_lattice_to_array
+import circuit
 from visual_array_cell import *
 
 from qiskit import circuit as qkcirc
@@ -13,18 +14,25 @@ GUISlice = List[List[Optional[VisualArrayCell]]] # 2D array of cells
 
 __all__ = ['compile_file','VisualArrayCell','GUISlice']
 
-def compile_file(circuit_file_name : str ,
-                 apply_litinski_transform:bool=True) -> Tuple[List[GUISlice], str]:
-    """Returns gui slices and the text of the circuit as processed in various stages"""
 
+def compile_file(circuit_file_name : str ,
+                 apply_litinski_transform:bool=True) :
+    """DEPRECATED. compile_str"""
+    with open(circuit_file_name) as input_file:
+        return compile_str(input_file.read(),apply_litinski_transform)
+
+
+def compile_str(qasm_circuit : str,
+                apply_litinski_transform:bool=True) -> Tuple[List[GUISlice], str]:
+    """Returns gui slices and the text of the circuit as processed in various stages"""
     composer_class = lattice_surgery_computation_composer.LatticeSurgeryComputation
     layout_types = lattice_surgery_computation_composer.LayoutType
 
-    input_circuit = segmented_qasm_parser.parse_file(circuit_file_name)
+    input_circuit = segmented_qasm_parser.parse_str(qasm_circuit)
 
     compilation_text = "Input Circuit:\n"
 
-    compilation_text += qkvis.circuit_drawer(qkcirc.QuantumCircuit.from_qasm_file(circuit_file_name)).single_string()
+    compilation_text += qkvis.circuit_drawer(qkcirc.QuantumCircuit.from_qasm_str(qasm_circuit)).single_string()
 
     compilation_text += "\nCircuit as Pauli rotations:\n"
     compilation_text += input_circuit.render_ascii()
--- a/lattice_surgery_compilation_pipeline.py
+++ b/lattice_surgery_compilation_pipeline.py
@@ -16,7 +16,7 @@ __all__ = ['compile_file','VisualArrayCell','GUISlice']
 
 
 def compile_file(circuit_file_name : str ,
-                 apply_litinski_transform:bool=True) :
+                 apply_litinski_transform:bool=True)  -> Tuple[List[GUISlice], str]:
     """DEPRECATED. compile_str"""
     with open(circuit_file_name) as input_file:
         return compile_str(input_file.read(),apply_litinski_transform)
--- a/webapi/json_pipeline.py
+++ b/webapi/json_pipeline.py
@@ -1,0 +1,45 @@
+import lattice_surgery_compilation_pipeline
+import json
+from typing import *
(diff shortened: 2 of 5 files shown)


webapi/local_compiler_api_server.py
```diff
@@ -0,0 +1,77 @@
...
+def probe(request_post_params: Optional[Dict] = None):
+    if request_post_params is None:
+        request_post_params = JSON_POST_PARAMETERS
+
+    conn = http.client.HTTPConnection('localhost:9876')
+    headers = {'Content-type': 'application/json'}
+
+    conn.request('POST', '/compile', json.dumps(request_post_params), headers)
+
+    response = conn.getresponse()
```
gwwatkin, line 70: How to call the API in python.

Alternatively one can debug by starting a Python console in project root and running:

`>>> import webapi.local_compiler_api_server`

`>>> webapi.local_compiler_api_server.probe()`

gwwatkin: This works even when `python webapi/local_compiler_api_server.py` is running in another console.

lattice_surgery_compilation_pipeline.py
```diff
@@ -13,18 +14,25 @@
 
 __all__ = ['compile_file','VisualArrayCell','GUISlice']
 
+
 def compile_file(circuit_file_name : str ,
-                 apply_litinski_transform:bool=True) -> Tuple[List[GUISlice], str]:
-    """Returns gui slices and the text of the circuit as processed in various stages"""
+                 apply_litinski_transform:bool=True) :
```
alexnguyenn:
```suggestion
                 apply_litinski_transform:bool=True)  -> Tuple[List[GUISlice], str]:
```

webapi/json_pipeline.py
```diff
@@ -0,0 +1,45 @@
...
+            obj_with_good_keys =  obj.__dict__
+            obj_with_good_keys['edges'] = dict([(k.value, v) for k, v in obj.edges.items()])
+            print(obj_with_good_keys)
+            return obj_with_good_keys
+        elif(isinstance(obj,object)):
+            return obj.__dict__
+        return obj
+
+
+def handle(json_request: str) -> JsonResponse:
```
alexnguyenn, line 33: Add docstrings to `handle()`

webapi/json_pipeline.py
```diff
@@ -0,0 +1,45 @@
...
+from typing import *
+import visual_array_cell
+import enum
+
+class JsonResponse:
+    def __init__(self, status: int, body: str):
+        self.status = status
+        self.body = body
+
+class SliceArrayJSONEncoder(json.JSONEncoder):
```
alexnguyenn: From discussion: make `SliceArrayJSONEncoder` private.

alexnguyenn (approved):
From discussion: add docstrings to `handle()` function and make `SliceArrayJSONEncoder` private, both

gwwatkin:
@alexnguyenn Addressed comments and merging

PR #132, opened 2021-11-02
Enable CORS in local api development server

gwwatkin:
Needed because of the different port of the web UI


alexnguyenn:
Did you add anything to `requirements.txt`?

gwwatkin:
> Did you add anything to `requirements.txt`?

Yes I did, good point. I added jsonpickle. Should have probably split the reformat commit and then adding the package

gwwatkin:
Also just noticed that there are changes made by the `black` formatter

alexnguyenn (approved):

alexnguyenn:
> > Did you add anything to `requirements.txt`?
> 
> Yes I did, good point. I added jsonpickle. Should have probably split the reformat commit and then adding the package

Yeah just need to keep track of that for setting up .cfg file for our project (though `jsonpickle` would go to dev part I think 🤔)

Issue #120, opened 2021-10-23
Add proper setup configuration


alexnguyenn:
To add `pyproject.toml` and `setup.cfg`.

alexnguyenn:
Right now our dependencies are `qiskit`, `pyzx`, `igraph` and `mako`. 
For dev we will add `black`, `flake8`, `isort`, `mypy`, `pytest` and `jsonpickle`.

PR #134, opened 2021-11-05
Adding configuration files 

alexnguyenn:
#120:
Added `.flake8`, `pyproject.toml` (which also contains configuration for black and isort) and `setup.cfg`. 

What still needs to be done:
* Fill out the missing `metadata` fields in `setup.cfg`
* Add `mypy` and `pytest` configurations. 
* Check and correct stuffs (especially dependencies)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,42 @@
+[metadata]
+name = lsqecc
+version = 
+description = a Python package for ... 
+long_description = file: README.md
+long_description_content_type = text/markdown; charset=UTF-8
+url = https://latticesurgery.com/
+author = George Watkins, Alex Nguyen, Varun Seshardi, Keeland Watkins
+author_email = 
+license = 
+license_file = 
+classifiers =
+    Topic :: Scientific/Engineering
+    Development Status :: 3 - Alpha
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Operating System :: OS Independent,
+    Intended Audience :: Science/Research
+    Natural Language :: English
+    Typing :: Typed
+project_urls = 
+    Documentation = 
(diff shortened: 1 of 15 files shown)


setup.cfg
```diff
@@ -0,0 +1,42 @@
...
+url = https://latticesurgery.com/
+author = George Watkins, Alex Nguyen, Varun Seshardi, Keeland Watkins
+author_email = 
+license = 
+license_file = 
+classifiers =
+    Topic :: Scientific/Engineering
+    Development Status :: 3 - Alpha
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
```
gwwatkin, line 16: Just curious, have you tried running with 3.8 and 3.7. I've only tried 3.9

setup.cfg
```diff
@@ -0,0 +1,42 @@
...
+packages = find:
+python_requires = >=3.7
+install_requires =
+    qiskit>=0.31.0
+    pyzx>=0.6.4
+    igraph>=0.9.8
+    mako>=1.1.5
+
+[options.extras_require]
+dev =
```
gwwatkin, line 37: Guess `mypy` will go here as well

gwwatkin (approved):

setup.cfg
alexnguyenn: Right now I am using 3.9 but I think I used it with 3.7 and 3.8 before. Might need to check again

setup.cfg
alexnguyenn: Yep, but right now it depends on our progress in [mypy-refactor](https://github.com/latticesurgery-com/lattice-surgery-compiler/tree/mypy-refactor). I don't want to have `mypy` blocking the whole thing.

gwwatkin:
Is this closed to be rebased on the latest branch?

PR #135, opened 2021-11-05
ops.pop() and remove_y_operators()

isolatedinformation:
added circuit_has_measurements() to detect Measurements blocks in Ciccuit

improvments in remove_y_operators() and apply_transformation()

### DIFF
--- a/lsqecc/pauli_rotations/circuit.py
+++ b/lsqecc/pauli_rotations/circuit.py
@@ -85,7 +85,7 @@ class PauliOpCircuit(object):
 
         """
         quarter_rotation = list()
-
+        circuit_has_measurements = self.circuit_has_measurements()
         # Build a stack of pi/4 rotations
 
         for i in range(start_index, len(self)):
@@ -102,7 +102,7 @@ class PauliOpCircuit(object):
             while index + 1 < len(self):
                 self.commute_pi_over_four_rotation(index)
                 index += 1
-            if self.circuit_has_measurements():
+            if circuit_has_measurements:
                 self.ops.pop()
 
     def remove_y_operators_from_circuit(self, start_index: int = 0) -> None:
@@ -112,7 +112,7 @@ class PauliOpCircuit(object):
 
         """
         i = start_index
-
+        circuit_has_measurements = self.circuit_has_measurements()
         while i < len(self.ops):
             pauli_block = self.ops[i]
             if isinstance(pauli_block, Measurement) or (
@@ -155,7 +155,7 @@ class PauliOpCircuit(object):
                         while right_block_index + 1 < len(self.ops):
                             self.commute_pi_over_four_rotation(right_block_index)
                             right_block_index += 1
-                        if self.circuit_has_measurements():
+                        if circuit_has_measurements():
                             self.ops.pop()
             else:
                 # This is assuming pi/4 rotations (not including ones from this operation)
--- a/lsqecc/pauli_rotations/circuit.py
+++ b/lsqecc/pauli_rotations/circuit.py
@@ -37,7 +37,9 @@ class PauliOpCircuit(object):
     def copy(self) -> "PauliOpCircuit":
         return copy.deepcopy(self)
 
-    def add_pauli_block(self, new_block: PauliProductOperation, index: int = None) -> None:
+    def add_pauli_block(
+        self, new_block: PauliProductOperation, index: int = None
+    ) -> None:
         """
         Add a rotation to the circuit
 
@@ -56,7 +58,11 @@ class PauliOpCircuit(object):
         self.ops.insert(index, new_block)
 
     def add_single_operator(
-        self, qubit: int, operator_type: PauliOperator, rotation_amount: Fraction, index: int = None
+        self,
+        qubit: int,
+        operator_type: PauliOperator,
+        rotation_amount: Fraction,
+        index: int = None,
     ) -> None:
         """
         Add a single Pauli operator (I, X, Z, Y) to the circuit.
@@ -75,7 +81,9 @@ class PauliOpCircuit(object):
 
         self.add_pauli_block(new_rotation, index)
 
-    def apply_transformation(self, start_index: int = 0, remove_y_operators: bool = True) -> None:
+    def apply_transformation(
+        self, start_index: int = 0, remove_y_operators: bool = True
+    ) -> None:
         """
         Apply Litinski's Transformation
 
@@ -85,7 +93,9 @@ class PauliOpCircuit(object):
         # Build a stack of pi/4 rotations
 
         for i in range(start_index, len(self)):
-            if isinstance(self.ops[i], PauliRotation) and self.ops[i].rotation_amount in {
+            if isinstance(self.ops[i], PauliRotation) and self.ops[
+                i
+            ].rotation_amount in {
                 Fraction(1, 4),
                 Fraction(-1, 4),
             }:
@@ -98,10 +108,11 @@ class PauliOpCircuit(object):
             while index + 1 < len(self):
                 self.commute_pi_over_four_rotation(index)
                 index += 1
-            self.ops.pop()
+            if self.circuit_has_measurements():
+                self.ops.pop()
 
-        if remove_y_operators:
-            self.remove_y_operators_from_circuit()
+        # if remove_y_operators:
+        #     self.remove_y_operators_from_circuit()
 
     def remove_y_operators_from_circuit(self, start_index: int = 0) -> None:
         """
@@ -113,7 +124,6 @@ class PauliOpCircuit(object):
 
         while i < len(self.ops):
             pauli_block = self.ops[i]
-
             if isinstance(pauli_block, Measurement) or (
                 pauli_block.rotation_amount in {Fraction(1, 8), Fraction(-1, 8)}
             ):
@@ -132,7 +142,9 @@ class PauliOpCircuit(object):
                     # rotations (one on each side)
                     if len(y_op_indices) % 2 == 0:
                         first_operator = y_op_indices.pop(0)
-                        self.add_single_operator(first_operator, PauliOperator.Z, Fraction(1, 4), i)
+                        self.add_single_operator(
+                            first_operator, PauliOperator.Z, Fraction(1, 4), i
+                        )
                         self.add_single_operator(
                             first_operator, PauliOperator.Z, Fraction(-1, 4), i + 2
                         )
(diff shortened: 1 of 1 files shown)


alexnguyenn:
did you set `line-length` to be 100 for black? Seems like there were a lot of changes made by black.

[here is the guide for it.](https://dev.to/adamlombard/vscode-setting-line-lengths-in-the-black-python-code-formatter-1g62)

lsqecc/pauli_rotations/circuit.py
```diff
@@ -75,7 +81,9 @@ def add_single_operator(
 
         self.add_pauli_block(new_rotation, index)
 
-    def apply_transformation(self, start_index: int = 0, remove_y_operators: bool = True) -> None:
+    def apply_transformation(
+        self, start_index: int = 0, remove_y_operators: bool = True
```
gwwatkin:
```suggestion
        self, start_index: int = 0
```
Recommend removing the unused flag

lsqecc/pauli_rotations/circuit.py
```diff
@@ -98,10 +108,11 @@ def apply_transformation(self, start_index: int = 0, remove_y_operators: bool =
...
                 self.commute_pi_over_four_rotation(index)
                 index += 1
-            self.ops.pop()
+            if self.circuit_has_measurements():
+                self.ops.pop()
 
-        if remove_y_operators:
-            self.remove_y_operators_from_circuit()
+        # if remove_y_operators:
+        #     self.remove_y_operators_from_circuit()
```
gwwatkin: Recommend removing the commented out lines

lsqecc/pauli_rotations/circuit.py
alexnguyenn: Also we would need to check in the package if anything is calling  `apply_transformation` with `remove_y_operators` as well.

lsqecc/pauli_rotations/circuit.py
```diff
@@ -198,7 +215,9 @@ def commute_pi_over_four_rotation(self, index: int) -> None:
             product_of_coefficients /= 1j
             if isinstance(self.ops[next_block], Measurement):
                 if product_of_coefficients.real < 0:
-                    self.ops[next_block].isNegative = not self.ops[next_block].isNegative
+                    self.ops[next_block].isNegative = not self.ops[
+                        next_block
+                    ].isNegative
```
gwwatkin: It looks like the lines are shrinking a little ... are you runin `black --line-length 100`?

This is what I think it was. This file should have already been formatted. Right @alexnguyenn?

isolatedinformation: What is our default line length? I had not set up that property. I thought the shrinking was part of the black's formatting. I have now set the line length to 100 in my editor

alexnguyenn: @gwwatkin: yep I mentioned it above, I think we would have to add another commit to revert the changes?

isolatedinformation: I am making another commit now. With suggestions on the same branch. Shoulf have a pr up in the next few mins

lsqecc/pauli_rotations/circuit.py
```diff
@@ -154,7 +155,8 @@ def remove_y_operators_from_circuit(self, start_index: int = 0) -> None:
                         while right_block_index + 1 < len(self.ops):
                             self.commute_pi_over_four_rotation(right_block_index)
                             right_block_index += 1
-                        self.ops.pop()
+                        if self.circuit_has_measurements():
```
alexnguyenn: Since `self.circuit_has_measurements()` is a loop itself, it is better to call it outside of the loop (at the top) and assign the result to a variable.

alexnguyenn (approved):

lsqecc/pauli_rotations/circuit.py
```diff
@@ -98,10 +102,8 @@ def apply_transformation(self, start_index: int = 0, remove_y_operators: bool =
             while index + 1 < len(self):
                 self.commute_pi_over_four_rotation(index)
                 index += 1
-            self.ops.pop()
-
-        if remove_y_operators:
-            self.remove_y_operators_from_circuit()
+            if circuit_has_measurements:
+                self.ops.pop()
```
gwwatkin, line 106: Sorry if I didn't catch this the first time around, but why do we pop only if the circuit has measurements? The way `self.circuit_has_measurements()` these measurements could be anywhere, not necessarily at the end.

I think this concept needs to be expressed better. The idea is that a rotation is past every measurement, we can pop it because it doesn't affect the probability distributions of the measurement readouts.

This however causes the transform not to be final state vector preserving.

gwwatkin: I think this is part of a deeper issue with our compiler. What are we trying to preserve? state vector correctness or readout distributions. If we make operations disappear we are no longer preserving state vector correctness with these transformation, but only measurement readout distributions...

@isolatedinformation I'd be very interested in your ideas on the matter

alexnguyenn: @gwwatkin Good question. I think `apply_transformation()` was developed based the assumption that the input circuit have measurements at the end. We may have to take a look at this again (maybe a new issue for it).

For the second point- as this behavior is restricted only to `apply_transformation()`, which the user can choose to turn on or off. So I am not sure if it is an compiler's issue 🤔; but it does add to the discussion of non-transformation vs transformation though.

isolatedinformation:
1. If we take our assumption that measurements occur at the end of the circuit, this implementation works because the pi/4 rotation gets absorbed into the measurement(correct me if I am wrong). In that case, would the state vector also not be preserved? 

2. `circuit_has_measurements()` and `pop()` is a very ad-hoc implementation. It can definitely be improved. 
image
This image shows that measurements are not always at the end of a circuit. But here the qubits after measurements start fresh and are disjoint from the measurements. In that case, I think our assumption is not wrong, 

3. I think it would be wise to also consider the case where the qubit state after the measurement is not disjoint from the measurement. In that case, the state of the qubit would be the post-measurement which can easily be calculated mathematically, if we know the state-vector or the density matrix before measurement. However to transfer the results of the measurement we would need to implement a classical register to transfer the measurement outcome.

lsqecc/pauli_rotations/circuit.py
isolatedinformation: 1. We are not making operations disappear. But rather using Litinski's transform Fig4(c) where the pi/4 measurements get absorbed. But I agree that pop() seems like a very ad-hoc implementation. 

2. **Improvement Possibilities**
image

The above image seems like a good start point to consider how measurements work in an actual quantum circuit. 
The measurement **do not** only appear at the end. But according to this image atleast, after each measurement the qubit line terminates and starts afr
After each Bell Measurements, in this circuit at least, the qubits are reinitialized (start afresh). This is an abstract circuit. It depends on the particular circuit in what state those qubits are present and there are two ways to do this(imo):
a. We continue with the Post Measurement State. 
This would require classical registers, to transfer the measurement information and then continue on the qubit line to
b. We initialize the state to some random state depending on the requirement of the circuit.
This would add more operations to the circuit.If we take our assumption that measurements occur at the end of the circuit, this implementation works because the pi/4 rotation gets absorbed into the measurement(correct me if I am wrong). In that case, would the state vector also not be preserved?

circuit_has_measurements() and pop() is a very ad-hoc implementation. It can definitely be improved.
image
This image shows that measurements are not always at the end of a circuit. But here the qubits after measurements start fresh and are disjoint from the measurements. In that case, I think our assumption is not wrong,

I think it would be wise to also consider the case where the qubit state after the measurement is not disjoint from the measurement. In that case, the state of the qubit would be the post-measurement which can easily be calculated mathematically, if we know the state-vector or the density matrix before measurement. However to transfer the results of the measurement we would need to implement a classical register to transfer the measurement outcome.

isolatedinformation: 1. If we take our assumption that measurements occur at the end of the circuit, this implementation works because the pi/4 rotation gets absorbed into the measurement(correct me if I am wrong). In that case, would the state vector also not be preserved?

2. `circuit_has_measurements()` and `pop()` is a very ad-hoc implementation. It can definitely be improved.
image
This image shows that measurements are not always at the end of a circuit. But here the qubits after measurements start fresh and are disjoint from the measurements. In that case, I think our assumption is not wrong,

3. I think it would be wise to also consider the case where the qubit state after the measurement is not disjoint from the measurement. In that case, the state of the qubit would be the post-measurement which can easily be calculated mathematically, if we know the state-vector or the density matrix before measurement. However to transfer the results of the measurement we would need to implement a classical register to transfer the measurement outcome.

isolatedinformation: 4. We would also need to come up with our own test cases for a circuit like the above. Because I have not come across examples of such a circuit.

gwwatkin: Thanks @alexnguyenn  and @isolatedinformation  for the super thoughtful replies.

@alexnguyenn 
 1. I think so. `apply_transformation()` was indeed developed with the idea that measurements would go at the end, and for that case we are on the same page as Litinski's GoSC, Pauli rotations after the last measurement don't matter. 
 2. I think you are right ... the user just needs to be aware of this aspect of toggling the Litinski transform. For example consider this case: we are just compiling a section of of a bigger circuit, in this case we want to preserve the output statevector. In this case it suffices to not enable the Litinski Transform, but this problem of being state vector preserving is something to be aware I think.
 
 @isolatedinformation 
 1. "Would the state vector also not be preserved" My understating is that it isn't, because the rotations after measurements are still making changes to the state vector - just that these changes have no bearing on the measurement readouts. I think this only matters when compiling a circuit that is part of a bigger quantum program, where the state might manipulated further by something else :shrug:
 2. Yep our assumption is not wrong ... as long as all we care about is only preserving readout distributions.
 3. What you call "the case where the qubit state after the measurement is not disjoint from the measurement" I think is what I call the state vector preserving case (or density matrix preserving). I suspect that all we need to do to make that case work is not pop operations anymore. We already have al the calculations that get the state after measurement, and yes, it would be nice to store the readout bit in a classical register:
 https://github.com/latticesurgery-com/lattice-surgery-compiler/blob/222e7f99ba2fec510bd125fb1ecbabda00af434e/lsqecc/simulation/logical_patch_state_simulation.py#L72-L75
 Two cases where we pop (now conditional up on whether we have measurements):
    - Litinski Transform
    - RHS corrective terms of Y removal commuted to the end.

Now I'm under the impression that we might want to change the criterion on whether we pop operations from having measurement to a user set flag :thinking: 
 4. Yep might want to add a note to the testing issue :innocent:

gwwatkin: > Now I'm under the impression that we might want to change the criterion on whether we pop operations from having measurement to a user set flag

This would also solve the problem I'm having now of some computations growing too large and crashing my computer if there are no measurements

gwwatkin: One more thing in favor of preserving the final state is that it is much easier to check for correctness of the compiled program :sweat_smile:

Issue #117, opened 2021-10-23
Adding Contribution Guide


alexnguyenn:
See: https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors

gwwatkin:
Things to add:
  * `feature_branch -> dev -> master/main` workflow
  * One approval required
  * [...]

gwwatkin:
NOTE: Remember to move the dev setup guide here, from the README

alexnguyenn:
@gwwatkin I think you can take on this since you already wrote the contribution part in our `README`.

[Here is an example](https://github.com/unitaryfund/mitiq/blob/master/CONTRIBUTING.md) from UF's `mitiq`, but there are many templates available online that we could use.

alexnguyenn:
Completed with #154

PR #143, opened 2021-11-09
Integrate `pre-commit` 

### LINKED ISSUE #124: Add `pre-commit` hook
See: https://pre-commit.com/

alexnguyenn:
Closes #124 

How to run:
* `pip install pre-commit` and then `pre-commit install` 
* To run: `pre-commit run` or just `git commit ...`, the first time running will take a while since it has to clone all the repos specified in `pre-commit-config.yaml`
* If want to run against all files (good for first run to test things out): `pre-commit run -a`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,36 @@
+# See https://pre-commit.com for more information
+
+default_language_version:
+  python: python3.9
+
+default_stages: [commit, push]
+exclude: ^debug/
+
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.0.1
+    hooks:
+      - id: check-case-conflict
+      - id: name-tests-test
+
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.10.0
+    hooks:
+      - id: isort
+        args: ["--check"]
+
+  - repo: https://github.com/ambv/black
+    rev: 21.10b0
+    hooks:
+      - id: black
+        args: ["--check"]
+
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 4.0.1
+    hooks:
+      - id: flake8
+
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.910
+    hooks:
+      - id: mypy
--- a/README.md
+++ b/README.md
@@ -47,6 +47,17 @@ Then install the package in edit mode:
 
 Now you can use this venv to work the package, from the command line and in an IDE. Make sure to select the venv you just created as the python environment for the project in your IDE.
 
+#### Pre commit hooks
+ Pre commit-hooks run type and style checks before every commit. Useful to avoid having to wait for feedback from CI and taking som load off of our pipelines.
+
+Set up instructions:
+* `pip install pre-commit` and then `pre-commit install` 
+* To run: `pre-commit run` or just `git commit ...`, the first time running will take a while since it has to clone all the repos specified in `pre-commit-config.yaml`
+
+If want to run against all files (good for first run to test things out): `pre-commit run -a`. 
+
+
+
 ## References
 <a id="1">[1]</a> 
 Alexandru Paler and Austin G Fowler. 


alexnguyenn:
@gwwatkin with `pre-commit install`, I am wondering whether or not we need `mypy` and `flake8` under `setup.cfg`. If a contributor don't use `pre-commit`, we can just do the checks via Github Actions once they submit a PR. 

Same case can be made for `isort` and `black`, but it is harder to decide since they are used as formatters. Right now `pre-commit run` runs them with `--check` so they just check if the code meets their specifications before a commit is made.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 4.0.1
+    hooks:
+      - id: flake8
+
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.910
+    hooks:
+      - id: mypy
```
gwwatkin, line 36: Shouldn't there be a `--ignore-missing-imports` in the args? can't test right now though

.pre-commit-config.yaml
```diff
@@ -0,0 +1,36 @@
...
+    rev: 5.10.0
+    hooks:
+      - id: isort
+        args: ["--check"]
+
+  - repo: https://github.com/ambv/black
+    rev: 21.10b0
+    hooks:
+      - id: black
+        args: ["--check"]
```
gwwatkin, line 26: Shouldn't we specify the line length here?

gwwatkin:
> @gwwatkin with `pre-commit install`, I am wondering whether or not we need `mypy` and `flake8` under `setup.cfg`. If a contributor don't use `pre-commit`, we can just do the checks via Github Actions once they submit a PR.
> 
> Same case can be made for `isort` and `black`, but it is harder to decide since they are used as formatters. Right now `pre-commit run` runs them with `--check` so they just check if the code meets their specifications before a commit is made.

My understanding was that if they are in the dev dependencies (which is the case) then people with the dev setup will have them... and yes, I would also expect the regular actions CI to catch errors if people don't have the hooks installed.

However I would add the line about the pre-commit set up in the "Dev setup" section of the readme

.pre-commit-config.yaml
alexnguyenn: No need since we already specified it in `pyproject.toml`

.pre-commit-config.yaml
alexnguyenn: same reason above

alexnguyenn:
Right now my understanding is `pre-commit` clone the tools you use from the repos you provide in `.pre-commit-config.yaml`, not from pypi. So there is no reason for contributors who are working with `pre-commit` to install `develop` dependencies as it will be redundant.

gwwatkin:
> Right now my understanding is `pre-commit` clone the tools you use from the repos you provide in `.pre-commit-config.yaml`, not from pypi. So there is no reason for contributors who are working with `pre-commit` to install `develop` dependencies as it will be redundant.

Yes I noticed that when setting up ... then do we even really need the dev dependencies on these 4 pre commit hook packages?

gwwatkin (approved):
Looks good :+1: Added the set up instructions to the README

alexnguyenn:
> Yes I noticed that when setting up ... then do we even really need the dev dependencies on these 4 pre commit hook packages?

Yeah that is my gripe. For people with `pre-commit` is redundant, but leaving those dependencies out for people without `pre-commit` means that they might not be aware that we are using those tools in development 😅 

I see you have been adding the contribution and set-up guide for `README.md`. We might want to move that under `CONTRIBUTION.md` for #117

gwwatkin:
> Yeah that is my gripe. For people with `pre-commit` is redundant, but leaving those dependencies out for people without `pre-commit` means that they might not be aware that we are using those tools in development 😅 

:+1:  good thinking ... Also they are such light packages it doesn't really matter.

> I see you have been adding the contribution and set-up guide for `README.md`. We might want to move that under `CONTRIBUTION.md` for #117

:ok_hand:  left a [note on the issue](https://github.com/latticesurgery-com/lattice-surgery-compiler/issues/117#issuecomment-964685928), for when we set it up properly.

gwwatkin:
I think we are good to merge

PR #145, opened 2021-11-11
Relocate compiler code to `src/lsqecc` and add testing directory

alexnguyenn:
Changes:
* Updating our directory to follow `src/mypackage` structure to keep things separated (and in prepare for adding `tox` if we need to). This is inspired by [Pytest doc: Good Integration Practice](https://docs.pytest.org/en/6.2.x/goodpractices.html#goodpractices), [Packaging a python library - ionel](https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure) and [Automated Testing in Python with pytest, tox, and GitHub Actions - mCoding](https://www.youtube.com/watch?v=DhUpxWjOhME). Already tested out with `pip install -e .`
* Update `.gitignore` with a heavily modified version of the template provided by [gitignore.io](https://www.toptal.com/developers/gitignore).
* Adding `tests/` directory. All future test codes will be placed under this directory.


gwwatkin:
FYI @isolatedinformation , in case you have some work in progress

gwwatkin (approved):
:+1: 

`pip install -e .` should still work right

alexnguyenn:
Yep I tested it

Issue #140, opened 2021-11-08
Support of equality in Circuit and Lattice classes


isolatedinformation:
Writing tests involves `assert` statements. Right now, if we compare two objects for equality, it returns `False`. By default, Python object identifiers for comparison operations. The classes that need to support equality are:

in lsqeccc/pauli_rotations:
- [x] `circuit.PauliOpRotation`
- [x] `rotation.PauliProductOperation`
- [x] `rotation.PauliRotation`
- [x] `rotation.Measurement`

in lsqecc/logical_lattice_ops:
- [ ] `llops.LogicalLatticeOperation` and associated child classes
- [ ] `llops.LogicalLatticeComputation`

This [thread](https://stackoverflow.com/questions/390250/elegant-ways-to-support-equivalence-equality-in-python-classes) has insights on implementing equality features.

alexnguyenn:
I will handle `pauli_rotations/` stuffs but @gwwatkin can you look at `logical_latice_ops/`?

gwwatkin:
@alexnguyenn I can of course, but my understanding was that @isolatedinformation was going to handle this

alexnguyenn:
Added equalities for `PauliRotation` and `Measurement`. There is no need for `PauliOperator` (being an Enum) and `PauliProductOperation` (being an Abstract Base Class - although I will need to rewrite that class a little bit)

gwwatkin:
@isolatedinformation is writing tests he'll decide if llops still need equality

isolatedinformation:
1. Llops does not need equality. Because we have already verified in Circuit if the ops are equal. Since equality is already implement for Circuit and Rotation classes, and llops attributes are composed of objects from these any assertion that needs to be made can exploit these inbuilt equalities. 
2. I don't see the end-user working with the internal details of llops, except for instantiating a LogicalLatticeComputation object. Everything else is internal and a good chunk of llops tests have already been written without need for equality.

gwwatkin:
Sounds good, @isolatedinformation. I agree

PR #146, opened 2021-11-12
Add equalities for `PauliRotation` and `Measurement` 

alexnguyenn:
As title. Tests for the added equalities included.

Part of #140.


alexnguyenn:
Opps I have not added any configurations for pytest yet - you would need to use `pytest tests` to run the test.

tests/rotation_test.py
```diff
@@ -0,0 +1,87 @@
+# Copyright (C) 2020-2021 - George Watkins and Alex Nguyen
```
gwwatkin: I recommend, adding the  `pauli_rotations` part to the path, like `tests/pauli_rotations/rotation_test.py`

gwwatkin: Since we went as far as adding a `src` folder I wonder whether it's wort adding the package name to the path:
```tests/lsqecc/pauli_rotations/rotation_test.py```

gwwatkin (approved):
Just a comment, about making the paths match between tests and the package

tests/rotation_test.py
alexnguyenn: No I don't think we need to do `tests/lsqecc` since we only have 1 package to test. We use `src` to make sure that other tools don't confuse between the test package and the actual package.

I am aligned with you on making the paths matched between tests and the package. We should also consider where to put our integration tests too.

gwwatkin: Integration tests usually go in a separate repo. We'd probably have them set up so you can choose to run them against `api.latticesurgery.com` or a local server.

alexnguyenn:
@gwwatkin I updated the tests directories to match with the compiler source code

gwwatkin (approved):

PR #144, opened 2021-11-11
Testing pre-commit CI

alexnguyenn:
A bad PR to test whether pre-commit CI works properly

PR #147, opened 2021-11-12
Update `pre-commit` configurations

alexnguyenn:
Changes:
* Remove `name-test-test` hook (it checks if files in tests/ end with `_test.py`) - removing because it might interfere when we are using pytest Fixture functions (and I prefer `test_*.py` better). 
* Adding `check-docstring-first` which checks for instances of code being placed in front of docstring. Fix one instance where this occurs.
* Remove `--check` option for both black and isort - this means pre-commit will applies change to your code (would have to do `git add` again if that happens). This also allow pre-commit.ci tool to push changes generated by those formatters to a PR - which I have set to off by default. If you want that to happen, comment "pre-commit.ci autofix" under your PR (~~also I would need to test that out after this PR is merged~~ Tested on #144). 

Finally here is [a list of all out-of-the-box hooks](https://github.com/pre-commit/pre-commit-hooks/blob/master/README.md) available. Let me know if you want any of them to be implemented.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,6 +3,9 @@
 default_language_version:
   python: python3.9
 
+ci:
+  autofix_prs: false
+
 default_stages: [commit, push]
 exclude: ^debug/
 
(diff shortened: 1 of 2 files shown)


gwwatkin (approved):
:+1: Thanks for the detailed description

gwwatkin:
I'll take a good look at the style checks, in the mean time you can go ahead and merge this

alexnguyenn:
Yeah just an extra note:
* The only 2 tools that will format the code (and cause pre-commit.ci to push) are black and isort. [Black](https://github.com/psf/black) reformats the code to its own code style and [isort ](https://github.com/PyCQA/isort)sorts all of our imports.
* This mean that even pre-commit.ci pushes a fix to your PR, the check can still fail if mypy and flake don't like it (you see it with #144), but at least black and isort won't complain anymore.

PR #148, opened 2021-11-14
Setup automated testing and dependabot

### LINKED ISSUE #122: Implement Automate Testing
To be implemented after https://github.com/latticesurgery-com/lattice-surgery-compiler/issues/100.

alexnguyenn:
As title. ~~Work in progress~~. Closes #122 

Changes:
* Adding settings for dependabot. 
* Adding workflows for testing on PR and push to dev and master. We will later expand this - 1 item I am thinking is uploading code coverage to codecov. 
* Splitting dependencies - core dependencies are pinned and keep in `setup.cfg` (so it can be updated by dependabot which trigger automated testings). Linting tools dependencies will be managed by pre-commit.ci thus no longer needed in `setup.cfg`. Testing, code coverage, doc, etc. dependencies will be kept in `dev-dependencies.txt`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,18 +27,12 @@ packages =
     lsqecc
 python_requires = >=3.9
 install_requires =
-    qiskit>=0.24.0
-    pyzx>=0.6.4
-    igraph>=0.9.8
-    mako>=1.1.5
+    qiskit>=0.24.0, <0.32.0
+    pyzx>=0.6.4, <0.6.5
+    igraph>=0.9.8, <0.9.9
 package_dir = 
     =src
 
 [options.extras_require]
-develop =
-    pytest>=6.2.5
-    black>=21.10b0
-    flake8>=4.0.1
-    isort>=5.10.0
-    jsonpickle>=1.3
-    mypy>=0.910
+latex =
+    mako>=1.1.5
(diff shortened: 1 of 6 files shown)


setup.cfg
```diff
@@ -25,20 +25,14 @@ project_urls =
 [options]
 packages = 
     lsqecc
-python_requires = >=3.9
+python_requires = >=3.7
```
gwwatkin, line 28: I'm curious as to why we are adding support for python 3.7 and 3.8. Supporting more versions can become a big responsibility, it can generate a lot of work and restrict us from using new features.

I'd be very careful before adding support for older python version, and wouldn't do it without a good reason.

gwwatkin:
Great to see that tests are running. Also 30s is pretty fast :+1:

setup.cfg
gwwatkin: After all this is a research project, not a foundational enterprise tool :sweat_smile: so I think we should try not to create work for us that isn't really serving anyone.

This might change if we have clients though, but the way this project is a cuting edge, I think functionality and results are the driving force.

gwwatkin (approved):

setup.cfg
alexnguyenn: Our main version is still 3.9. Right now I am having that setting there to take advantage of Github Actions allowing us to run tests on different python version in parallel. Once problems arise 3.7 or 3.8 that we cannot resolve quickly we will remove support for that version.

gwwatkin (approved):

PR #149, opened 2021-11-14
PauliOpCircuit equality

isolatedinformation:
-  added equality and not equal  of PauliOpCircuit
- added tests for PauliOpCircuit equality

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,8 +69,8 @@ __pypackages__/
 # Environments
 .env
 .venv
-env/
-venv/
+env*/
+venv*/
 ENV/
 env.bak/
 venv.bak/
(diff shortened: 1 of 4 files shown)


gwwatkin (approved):
Should be good. I would wait for #148 to go in first so we can test the CI

alexnguyenn:
Just minor stuffs but I think we would need at least 3 test cases for each functions (as we have multiple checks in `__eq__`).

isolatedinformation:
Aren't there 3 test cases for both functions? I made three different circuits, all the things are checked for.

alexnguyenn:
oh mb I missed that. I see how it works now

PR #151, opened 2021-11-14
Test CI 

alexnguyenn:
Testing if caching works and other workflows behavior with our old `test-precommit-ci` branch


alexnguyenn:
Seems working fine to me, tests are not failing because we don't have tests haha

PR #153, opened 2021-11-15
Add documentation directory 

alexnguyenn:
As title. Documentation is located at `docs/`. We will be using [myst-parser](https://myst-parser.readthedocs.io/en/latest/index.html) which allow us to Markdown instead of reStructuredText, as well as re-using our existing `.md` file to the documentation page. ~~Also created a automated job for testing documentation~~ switched to using ReadTheDocs build tool. Work in progress.

Note - to build on your own, use this:
```
cd docs
pip install -r requirements.txt
make html
```

Part of #115.

PR #157, opened 2021-11-16
Add `pauli_rotations` to API documentation

alexnguyenn:
Changes:
* Add section for `pauli_rotations` submodule to demonstrate [Sphinx autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) usage. 
* Trying out ReadTheDocs theme - API page with Furo was hard to read for me. 
* Trying out [sphinx-autodoc-typehints](https://github.com/agronholm/sphinx-autodoc-typehints). Seems good so far. 

Also testing if CI tool skip testing for this PR.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,25 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-20.04
+  tools:
+    python: "3.9"
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+    builder: html
+    configuration: docs/source/conf.py
+    fail_on_warning: true
+
+# Optionally declare the Python requirements required to build your docs
+python:
+    install:
+      - requirements: docs/requirements.txt
+      - method: pip
+        path: .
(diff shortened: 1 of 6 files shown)


alexnguyenn:
@gwwatkin to continue our discussion from #153, this is how to use Sphinx's autodoc in our documentation. 

I also noticed that for RTD to build successfully using autodoc it has to install our package (+ all of the dependencies) which increases the build time significantly (to 2 minutes not including queue time).

gwwatkin (approved):
The docs for the Pauli Rotations look really nice, great stuff :sunglasses:. I'd be inclined to add docs like this for every file ... do you know of a way to populate the API Reference page automatically with a file hierarchy?

I think the 2 min time is fine unless it causes our queue times to be much worse. The CI pipelines are just a check right, people should have their changes pre-tested/linted anyways.

gwwatkin:
I'd call a <5 min pipeline pretty quick

alexnguyenn:
Running into problems with RTD build and autodoc not being able to find the project so I am fixing it now.

@gwwatkin [sphinx-apidoc](https://www.sphinx-doc.org/en/master/man/sphinx-apidoc.html) was the tool I used. However it was very messy and I prefer doing it manually. 

Here is how to call it (at root):
```
sphinx-apidoc -o docs/source/api src/
```
You would have to take the `toctree` that include all the submodules at `api/lsqecc.rst` and throw them into `api/index.md` then remake the page.

alexnguyenn:
@gwwatkin Finally fixed the issue with RTD. It was not installing the package in editable mode, and I had to edit sys path in `source/conf.py` to point to it. autodoc should be working normally on there now: [See pauli_rotations on RTD](https://lattice-surgery-compiler--157.org.readthedocs.build/en/157/api/pauli_rotations.html). Also update settings for RTD to fail build when it encounters warnings (so we can catch it before deploying). 

Also an extra note: run `make clean` before `make html` to make sure you are getting a fresh build of the documentation (I was running into weird behaviors where the old build would affect the new build).

PR #160, opened 2021-11-16
Switch back to Furo theme for documentation

alexnguyenn:
As title. ~~Also testing if `.readthedocs.yaml` is really needed~~ yep still need it as of right now.


gwwatkin (approved):
Doc build looks good. Appreciate the fact you can toggle the dark theme

PR #163, opened 2021-11-18
[BUGFIX] Product of Pauli Operators was defined wrong.

isolatedinformation:
- Corrected the output of the prodcuts
- line 254 in circuit.py should be multiplication not division.
- made relevant changes to rotation.py tests

### DIFF
--- a/src/lsqecc/pauli_rotations/circuit.py
+++ b/src/lsqecc/pauli_rotations/circuit.py
@@ -224,7 +224,7 @@ class PauliOpCircuit(object):
 
             # Flip the phase if product of coefficients is negative
             # Product of coefficients will always be either i or -i (see issues #28 for proof)
-            product_of_coefficients /= 1j
+            product_of_coefficients *= 1j
             if isinstance(self.ops[next_block], Measurement):
                 if product_of_coefficients.real < 0:
                     measurement = cast(Measurement, self.ops[next_block])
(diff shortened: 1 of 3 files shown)


src/lsqecc/pauli_rotations/circuit.py
```diff
@@ -224,7 +224,7 @@ def commute_pi_over_four_rotation(self, index: int) -> None:
 
             # Flip the phase if product of coefficients is negative
             # Product of coefficients will always be either i or -i (see issues #28 for proof)
-            product_of_coefficients /= 1j
+            product_of_coefficients *= 1j
```
alexnguyenn, line 227: I am not 100% sure rn but I think the reason we use `*= 1j` is for checking whether we are getting i or -i.

src/lsqecc/pauli_rotations/rotation.py
```diff
@@ -78,10 +78,10 @@ def multiply_operators(a: "PauliOperator", b: "PauliOperator"):
 ] = {
     (PauliOperator.Z, PauliOperator.X): (1j, PauliOperator.Y),
     (PauliOperator.X, PauliOperator.Z): (-1j, PauliOperator.Y),
-    (PauliOperator.Z, PauliOperator.Y): (1j, PauliOperator.X),
-    (PauliOperator.Y, PauliOperator.Z): (-1j, PauliOperator.X),
-    (PauliOperator.Y, PauliOperator.X): (1j, PauliOperator.Z),
-    (PauliOperator.X, PauliOperator.Y): (-1j, PauliOperator.Z),
+    (PauliOperator.Y, PauliOperator.Z): (1j, PauliOperator.X),
```
alexnguyenn, line 81: how did I even miss this… 😱

gwwatkin (approved):
Works as agreed over zoom

src/lsqecc/pauli_rotations/rotation.py
gwwatkin: I think there was some confusion about what this table is, we need to comment with what is the invariant between the tuple elements:
`(A,B),(k,C) => A*B = k*C`

PR #165, opened 2021-11-19
Add test cases for `Measurement` and `PauliRotation`

alexnguyenn:
As title. Only missing `__str__()` and `to_latex()` which I am too lazy to touch right now 😅

PR #166, opened 2021-11-19
Change the implementation of `PauliProductOperation` 

alexnguyenn:
Made `PauliProductOperation` an [ABC class](https://docs.python.org/3.9/library/abc.html). `ConditionalOperation` mix-in is now used by `Measurement` and `PauliRotation` instead. 

~~Will need to rebase with dev to trigger automated testing again once #165 is approved and merged~~. Done

### DIFF
--- a/src/lsqecc/pauli_rotations/rotation.py
+++ b/src/lsqecc/pauli_rotations/rotation.py
@@ -91,12 +91,14 @@ class PauliProductOperation(ABC):
         self.qubit_num: int = no_of_qubit
         self.ops_list: List[PauliOperator] = [PauliOperator("I") for i in range(no_of_qubit)]
 
+    @abstractmethod
     def __str__(self) -> str:
         pass
 
     def __repr__(self) -> str:
         return str(self)
 
+    @abstractmethod
     def __eq__(self, other):
         pass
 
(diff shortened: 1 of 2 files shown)


gwwatkin (commented):
I'm ok with taking the conditional operator out of `PauliProductOperation`. But I'm not too sure about `change_single_op`,  `get_op`, `get_ops_map` needing to be abstract. There isn't much point if then we just forward them directly with the `super()` syntax.

On the other hand it might make sense to make `__str__` and `to_latex` abstract since their implementation might differ in the child class

alexnguyenn:
pre-commit.ci autofix

src/lsqecc/pauli_rotations/rotation.py
```diff
@@ -91,12 +91,14 @@ def __init__(self, no_of_qubit: int):
...
         self.ops_list: List[PauliOperator] = [PauliOperator("I") for i in range(no_of_qubit)]
 
+    @abstractmethod
     def __str__(self) -> str:
         pass
 
     def __repr__(self) -> str:
         return str(self)
 
+    @abstractmethod
```
gwwatkin, line 101: Yep this is 100% the right kind of abstract method, I also missed it before it seems

gwwatkin (approved):

PR #168, opened 2021-11-21
Add tests for `__str__` and `to_latex()` for Measurement and PauliRotation

### LINKED ISSUE #167: Add tests __str__ and to_latex tests for the Pauli rotation classes
Issue Description
-----------------

See #165

Proposed Solution
-----------------

Test these methods comparing outputs

alexnguyenn:
Closes #167. Also switch to using f-string for those functions.

### DIFF
--- a/src/lsqecc/pauli_rotations/rotation.py
+++ b/src/lsqecc/pauli_rotations/rotation.py
@@ -107,7 +107,7 @@ class PauliProductOperation(ABC):
 
     @abstractmethod
     def to_latex(self) -> str:
-        return_str = "(" + str(self.ops_list[0])
+        return_str = f"({self.ops_list[0]}"
         if self.qubit_num > 1:
             for i in range(1, len(self.ops_list)):
                 return_str += r" \otimes" + " " + str(self.ops_list[i])
@@ -167,7 +167,7 @@ class PauliRotation(PauliProductOperation, coc.ConditionalOperation):
         self.rotation_amount: Fraction = rotation_amount
 
     def __str__(self) -> str:
-        return "{}: {}".format(self.rotation_amount, self.ops_list)
+        return f"{self.rotation_amount}: {self.ops_list}"
 
     def __eq__(self, other) -> bool:
         return (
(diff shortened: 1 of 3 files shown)


tests/pauli_rotations/rotation_test.py
```diff
@@ -91,6 +91,37 @@ def test_pauli_rotation_eq(self, pauli_rotation_1, pauli_rotation_2):
...
+    )
+    def test_str(self, input, expected):
+        assert str(input) == expected
+
+    @pytest.mark.parametrize(
+        "input, expected",
+        [
+            (PauliRotation.from_list([Y], Fraction(1, 2)), r"(Y)_\frac{\pi}{2}"),
+            (PauliRotation.from_list([Z, I], Fraction(-0, 4)), r"(Z \otimes I)_0"),
+            (PauliRotation.from_list([X, Y], Fraction(-6, 3)), r"(X \otimes Y)_-2\pi"),
```
gwwatkin: On mobile, can't run the tests, but shouldn't shouldn't the index script be `_{-2*pi}` looks like we might have a bug

gwwatkin: Since the ci passes

gwwatkin (commented):
Away from my computer, but I see some strange index behavior bin the tests. It's probably due to some pre existing bug

tests/pauli_rotations/rotation_test.py
alexnguyenn: I just checked with overleaf. It should be `(X \otimes Y)_{-2\pi}` (which will produce something like below) so we are having a bug here.

image

alexnguyenn: Just pushed a fix

gwwatkin: I think its good now

src/lsqecc/pauli_rotations/rotation.py
```diff
@@ -208,7 +205,8 @@ def __init__(self, no_of_qubit: int, isNegative: bool = False) -> None:
         self.isNegative: bool = isNegative
 
     def __str__(self) -> str:
-        return "{}M: {}".format("-" if self.isNegative else "", self.ops_list)
+        sign = "-" if self.isNegative else ""
+        return f"{sign}M: {self.ops_list}"
```
gwwatkin, line 209: This does look cleaner

gwwatkin (approved):
:+1:

PR #169, opened 2021-11-21
Add automate labeling 

alexnguyenn:
See https://github.com/actions/labeler

Will need a test period to see if it works properly with our PRs.


gwwatkin (approved):
Nice feature. We can always add more labels as we go

Issue #71, opened 2021-04-02
Create dependency DAG of circuit of Pauli rotations operations


gwwatkin:
To be used later by a scheduler to parallelize operations on the lattice. 

Operations that act on different qubits can be parallelized. Need to look into how commuting plays here too. Perhaps things that commute, even when act on the same qubits, should be neighbors in the tree.

gwwatkin:
**Nice to have:** The tree is abstract so it can be reused for logical lattice operations, which sit between  the circuit of Pauli rotations and actual patch operations

gwwatkin:
**Idea:** Start from the back of the circuit (chronologically latest) and work bacwards.
Thinking about it more carefull what we need to construct id a dependency DAG, not a tree
**Example:**
```
Op name:(A)  (B)  (C)  (D)  (E)  (F)
    q1--|X|--|I|--|X|--|I|--|Z|--|I|
    q2--|I|--|Z|--|Z|--|X|--|I|--|Z|
   pi*  1/4  1/4  1/8  1/4   M    M 
```
Tree would look something like:
![DependancyDAG](https://user-images.githubusercontent.com/36427091/113698357-3e02f800-9689-11eb-8539-690aba92231a.png)

Where the thicker circled nodes are terminal and don't depend on anything - they are like the result.

The prototypes could look like so:
```
from typing import *
from circuit import *

T = TypeVar('T')

class DependancyTree: # static class/ namespace

    class Node:
        def __init__(self, op : T, dependencies : List['DependancyTree.Node[T]']):
            self.op = op
            self.dependencies = dependencies

    @staticmethod
    def from_circuit(circuit : Circuit) -> List['DependancyTree.Node[T]']:
    """Returns the list of terminal nodes"""
        pass
```

alexnguyenn:
Since some cases require traversing the tree backwards, we need to keep track of parents as well.

gwwatkin:
It would be interesting to look into the run time analysis of the  algorithm. Intuitively it looks like `O(depth^2)` but there are only so many pauli matrices, so maybe there is a nice counting argument for which we'll always run into a non commuting operator :thinking:

alexnguyenn:
Note: A and B are both commute to C, therefore the above diagram is incorrect.

gwwatkin:
Thanks! Updated the circuit.

PR #173, opened 2021-11-23
Add tests for DependencyGraph

### LINKED ISSUE #172: Test `dependency_graph.py`
As title. Expected to be painful since I don't recall how I implemented it.

alexnguyenn:
Closes #172. Work in progress.

### DIFF
--- a/src/lsqecc/logical_lattice_ops/dependency_graph.py
+++ b/src/lsqecc/logical_lattice_ops/dependency_graph.py
@@ -24,6 +24,8 @@ T = TypeVar("T")
 
 
 class DependencyGraph:
+    """Class representing a dependency DAG. Experimental"""
+
     class Node:
         def __init__(self, op: T):
             self.op = op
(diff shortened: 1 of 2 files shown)


alexnguyenn:
@gwwatkin: still have `from_list` and `from_circuit_by_commutation` left to write tests for. I am planning to use the example from #71 but any insight will be appreciated.

gwwatkin (approved):
@alexnguyenn I suggested in #172 putting this on pause, since it might be some time before we start using it. What's here is good. I propose we commit what we have and leave an "experimental" note in the docstring of the classes exported by this file.

If you want to proceed anyways, I think the case in 71 is good and I can help you come up with more. I'd be careful about writing tests for classes that are not involved yet in the main compilation. It might turn out that when we start to use the practically we'll have to make heavy changes.

gwwatkin:
If you choose to merge you can just move #172 to the backlog

alexnguyenn:
I think I will add the test case in #71 and then merge this - actually it might already be in `debug/` already.

gwwatkin:
Sounds like a good plan. I might just consider a note that says experimental to the doctring

alexnguyenn:
added the test case from #71 - I think I am done with this. Ran into some problems with PauliRotation being non-hashable (see #175).

PR #175, opened 2021-11-24
add `__hash__` for Measurement and PauliRotation

alexnguyenn:
Learned that since we override `__eq__`, we need to also provide our own implementation of `__hash__` if we want the class to be hashable (to be able to use in dict or set). The problem is both Measurement and PauliRotation are not meant to be mutable, so we might come into some troubles with this. 

Need more opinions on this @gwwatkin.

### DIFF
--- a/src/lsqecc/pauli_rotations/rotation.py
+++ b/src/lsqecc/pauli_rotations/rotation.py
@@ -177,7 +177,7 @@ class PauliRotation(PauliProductOperation, coc.ConditionalOperation):
         )
 
     def __hash__(self) -> int:
-        return hash((self.rotation_amount, tuple(self.ops_list)))
+        return hash(hash(self.rotation_amount) + hash(tuple(self.ops_list)))
 
     def to_latex(self) -> str:
         return f"{super().to_latex()}_{{{phase_frac_to_latex(self.rotation_amount)}}}"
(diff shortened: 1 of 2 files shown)


alexnguyenn:
After some reading I am not too sure about this. If we are inserting blocks from a circuit into a dictionary, I feel like 2 blocks even with similar ops and rotation amount should be considered unique blocks. But then it does not make sense for 2 equal objects to have different hash values 🤔.

gwwatkin:
@alexnguyenn I think it was just a matter of combining the hashes, does it make sense now?

gwwatkin:
The only constraint that matters is that `A==B   =>   hash(A)==hash(B)`, the converse `hash(A)==hash(B)   =>  A==B ` , doesn't need to hold, but data structures and algorithms work better the closer it is to holding.

So basically all that matters is being consitent with `__eq__`, for example:
```py
    def __eq__(self, other) -> bool:
        return (
            isinstance(other, Measurement)
            and self.isNegative == other.isNegative
            and self.ops_list == other.ops_list
        )
```
 where `self.ops_list == other.ops_list` turns out to actually do a deep check

gwwatkin:
Violating `A==B => hash(A)==hash(B)` will break things like `set` and `dict` in unexpected ways

alexnguyenn:
Update the tests with `pytest.mark.parametrize`. @gwwatkin will need your approval since I am the PR author.

Interestingly, I noticed that because Github Actions is checking out the merged source code (which was just updated with the tests from #173). This might be the reason why codecov is also including those changes in its report.

gwwatkin (approved):

Issue #156, opened 2021-11-15
[BUG] Commutations between PauliRotation and Measurement


isolatedinformation:
Issue Description
-----------------

The commutations rules from Litinski Fig 4. are implemented incorrectly for commutations between PauliRotation block and Measurement block.

How to Reproduce
----------------
Run `tests/pauli_rotations/circuit_test.py::test_apply_transformation()`
The third fixture should fail. 

Problems
-------------
- For the two cases: P and Q commute and P and Q anticommute:
    - [ ] how do exp(i*pi/4*P) and exp(i*pi/4*theta) commute? 
    - [ ] how do exp(i*pi/4*P) and M_{Q} commute? What are the readout distributions?
    - [ ] how do exp(i*pi/4*P) and M_{-Q} commute? What are the readout distributions?
- What about the negative cases like exp(-i*pi/4*P)?
    - [ ] Do we just use exp(-i*pi/4*P) = exp(i*pi*7/4*P)^7 = exp(i*pi/4*P)^7 (sequential application 7 times?)

To-Do
--------
- [ ] Break `commute_pi_over_four_rotations()` into small helper functions for different commutation rules
- [ ] Refactor `apply_transformation()` to adjust for the changes above
- [ ] Tests for the helper functions

gwwatkin:
For the negative cases
> Do we just use exp(-ipi/4P) = exp(ipi7/4P)^7 = exp(ipi/4*P)^7 (sequential application 7 times?)

It worked in the case when commuting with other rotations (from https://github.com/latticesurgery-com/lattice-surgery-compiler/pull/29#discussion_r584187612), see below:

**Note** that the anti commuting case still needs to be checked, but this approach might serve as inspiration.

![00336F56-A85A-4B27-98D3-1FB741B3EA35](https://user-images.githubusercontent.com/36427091/109399510-83722e00-7943-11eb-89bc-0a31b1da7d73.png)

gwwatkin:
Do the rules in (a) also apply to (b)? I.e. can we commute measurements the same way we commute Pauli rotations?
![Screenshot_20211118_101550](https://user-images.githubusercontent.com/36427091/142473214-938187b3-6e4b-485a-9f1d-b77b791df1da.png)

Litinski does it in his presentation.

![Screenshot_20211118_101638](https://user-images.githubusercontent.com/36427091/142473354-adf31662-ef4f-4fd8-9bcb-b76c8b03183a.png)

[...]

PR #177, opened 2021-11-28
Add test cases for commutation

alexnguyenn:
Work in progress. Currently missing test cases for commuting Pauli rotation over a measurement block. 

Need reviews for test cases.

Part of #156

### DIFF
--- a/tests/pauli_rotations/circuit_test.py
+++ b/tests/pauli_rotations/circuit_test.py
@@ -136,3 +136,20 @@ def test_commute_rotation_measurement_commuting(rotation, measurement):
 def test_commute_rotation_measurement_anticommute():
     # TODO: Develop test cases for this test
     pass
+
+
+def test_commute_no_next_block():
+    circuit = PauliOpCircuit(1)
+    circuit.add_pauli_block(PauliRotation.from_list([X], Fraction(1, 4)))
+
+    with pytest.raises(Exception):
+        circuit.commute_pi_over_four_rotation(0)
+
+
+def test_commute_non_pi_over_4_rotation():
+    circuit = PauliOpCircuit(1)
+    circuit.add_pauli_block(PauliRotation.from_list([X], Fraction(1, 8)))
+    circuit.add_pauli_block(PauliRotation.from_list([Y], Fraction(1, 4)))
+
+    with pytest.raises(Exception):
+        circuit.commute_pi_over_four_rotation(0)
--- a/tests/pauli_rotations/circuit_test.py
+++ b/tests/pauli_rotations/circuit_test.py
@@ -75,9 +75,12 @@ def test_count_rotations_by(circuit, fraction, expected):
     ],
 )
 def test_commute_rotation_to_rotation_commuting(rotation1, rotation2):
+    # Before commuting: rotation1 ---- rotation2
     test_circuit = PauliOpCircuit(len(rotation1[0]))
     test_circuit.add_pauli_block(PauliRotation.from_list(*rotation1))
     test_circuit.add_pauli_block(PauliRotation.from_list(*rotation2))
+
+    # After commuting: rotation2 ---- rotation1
     test_circuit.commute_pi_over_four_rotation(0)
 
     assert test_circuit.ops[0].ops_list == rotation2[0]
@@ -87,7 +90,7 @@ def test_commute_rotation_to_rotation_commuting(rotation1, rotation2):
 
 
 @pytest.mark.parametrize(
-    "rotation1, rotation2, resultant_rotation",
+    "rotation1, rotation2, rotation2_after",
     [
         (([X], Fraction(1, 4)), ([Z], Fraction(1, 4)), ([Y], Fraction(1, 4))),
         (([Y], Fraction(-1, 4)), ([X], Fraction(-1, 8)), ([Z], Fraction(-1, 8))),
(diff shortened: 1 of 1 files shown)


tests/pauli_rotations/circuit_test.py
```diff
@@ -57,6 +65,74 @@ def test_count_rotations_by(circuit, fraction, expected):
...
+    [
+        (([X], Fraction(1, 4)), ([Z], Fraction(1, 4)), ([Y], Fraction(1, 4))),
+        (([Y], Fraction(1, 4)), ([X], Fraction(-1, 8)), ([Z], Fraction(-1, 8))),
+        (([Z, X], Fraction(1, 4)), ([Y, X], Fraction(1, 8)), ([X, I], Fraction(1, 8))),
+        (([X, Z, Y], Fraction(1, 4)), ([Y, X, Z], Fraction(-1, 8)), ([Z, Y, X], Fraction(-1, 8))),
+        (
+            ([X, Y, X, Z, Z], Fraction(1, 4)),
+            ([Z, Y, I, X, Y], Fraction(1, 8)),
+            ([Y, I, X, Y, X], Fraction(1, 8)),
+        ),
```
alexnguyenn, line 104: For this and other anti-commuting cases, reviews are needed on the sign of the resultant rotation (third one). According to https://github.com/latticesurgery-com/lattice-surgery-compiler/issues/28#issuecomment-786611605, it should be -pi/8 for  line 99?

tests/pauli_rotations/circuit_test.py
```diff
@@ -57,6 +65,91 @@ def test_count_rotations_by(circuit, fraction, expected):
...
+    test_circuit.commute_pi_over_four_rotation(0)
+
+    assert test_circuit.ops[0].ops_list == rotation2[0]
+    assert test_circuit.ops[0].rotation_amount == rotation2[1]
+    assert test_circuit.ops[1].ops_list == rotation1[0]
+    assert test_circuit.ops[1].rotation_amount == rotation1[1]
+
+
+@pytest.mark.parametrize(
+    "rotation1, rotation2, resultant_rotation",
```
gwwatkin: I might suggest renaming `resultant_rotation` to something that makes it clear that it is what rotation2 becomes, so maybe `first_resulting_rotation` ? because now it looks like two rotations are becoming one, but that is not the case.

Then also possibly add a comment documenting the fact that the second resulting rotation stays unchanged. I see you are checking it in the test case but it's not immediately obvious
``````

File: tests/pauli_rotations/circuit_test.py [hunk 3]
```diff
@@ -57,6 +65,91 @@ def test_count_rotations_by(circuit, fraction, expected):
...
+    test_circuit.commute_pi_over_four_rotation(0)
+
+    assert test_circuit.ops[0].ops_list == measurement[0]
+    assert test_circuit.ops[0].isNegative == measurement[1]
+    assert test_circuit.ops[1].ops_list == rotation[0]
+    assert test_circuit.ops[1].rotation_amount == rotation[1]
+
+
+def test_commute_rotation_measurement_anticommute():
+    # TODO: Develop test cases for this test
```
gwwatkin: good to have Codecov here that is going to keep reminding of the unchecked branch. The proof will be ready soon

gwwatkin (commented):
Code-wise the tests look good, just added a minor note on documenting some test cases.

However, we might want to work the tests out by hand before merging? and maybe add a couple more parametrize entries? like separating commuting because of different parts of the circuit Eg (X otimes I) and (I otimes Z) and commuting because of identical ops? 

I would be happy to work on this myself but I don't have the bandwidth right now, @isolatedinformation do you think you could have a look?

tests/pauli_rotations/circuit_test.py
gwwatkin: It does look like the sign should be changing

gwwatkin: Another thing to fix before merging this

gwwatkin: I think the mistake is here:
https://github.com/latticesurgery-com/lattice-surgery-compiler/blob/d0c60f3299257837b963958e24955177eb5ddcf0/src/lsqecc/pauli_rotations/circuit.py#L233-L235

gwwatkin: but I still need to go through it properly

tests/pauli_rotations/circuit_test.py
alexnguyenn: Good point - I will add that in.

alexnguyenn:
@gwwatkin I forgot to make the PR as a draft 😅. But yeah we need to verify those test cases and add more if it is possible (parametrize makes this relatively easy). Feel free to make suggestions.

tests/pauli_rotations/circuit_test.py
isolatedinformation: The resultant rotation should have the fraction as +1/8. That is the correct answer.  If you do the calculations for iPQ, the coefficient will come out to be +1

isolatedinformation: There should also be a case to test where the sign of the `resulltant_rotation` gets flipped.. All the examples here retain the sign of `rotation_2`. For example, in line 95 exchanging one Pauli Opertor between the rotations should flip the sign of the `resultant_rotation`. Like below
```
        (([X, Z, Z], Fraction(1, 4)), ([Y, X, Y], Fraction(-1, 8)), ([Z, Y, X], Fraction(1, 8))),
```

tests/pauli_rotations/circuit_test.py
isolatedinformation: The same cases from rotation, rotation anticommuting should work here. We have the proof that when they anticoommute the measurement and the Pauli Rotation commute

isolatedinformation (commented):
I added some changes for the anticommuting case as well as test cases for the anticommutation between a Rotation and Measurement block

tests/pauli_rotations/circuit_test.py
gwwatkin: What about the calculations in 28, they make it look like the sign should be negative?

gwwatkin: Ok I see the mistake
Screenshot_20211201_121720-1

gwwatkin: so the problem is with the example not the code

gwwatkin (commented):
@isolatedinformation Thanks for adding more tests

isolatedinformation (approved):
I have worked out the cases by hand for all the commutations tests. They are okay from my side.

tests/pauli_rotations/circuit_test.py
gwwatkin: Yes, we do have the proof now

gwwatkin (approved):
I trust @isolatedinformation  for the correctness of the calculations

Happy with the coverage of the commutation code. If with more use we find some particular cases that we missed we have all the structure to add them in quickly. Very happy with these changes.

There still are some things to test, such as `are_commuting` which this code relies on, though they can be addressed in another pull request, I'd rather see these changes in sooner.

tests/pauli_rotations/circuit_test.py
alexnguyenn: comments added

PR #179, opened 2021-12-02
refactoring commute_pi_over_four_rotation

isolatedinformation:
- The function `commute_pi_over_four_rotation` is redefined as `swap_adajcent_blocks`
- This method calls two private methods `__swap_adjacent_commuting_blocks` and `__swap_adjacent_anticommuting_blocks`

### DIFF
--- a/src/lsqecc/pauli_rotations/circuit.py
+++ b/src/lsqecc/pauli_rotations/circuit.py
@@ -191,7 +191,7 @@ class PauliOpCircuit(object):
 
             i += 1
 
-    def __swap_adjacent_commuting_blocks(self, index: int) -> None:
+    def _swap_adjacent_commuting_blocks(self, index: int) -> None:
         """
         Move a pi over four rotation block past its' neighbor block when the blocks commute
 
@@ -203,7 +203,7 @@ class PauliOpCircuit(object):
         self.ops[index] = self.ops[next_block]
         self.ops[next_block] = temp
 
-    def __swap_adjacent_anticommuting_blocks(self, index: int) -> None:
+    def _swap_adjacent_anticommuting_blocks(self, index: int) -> None:
         """
         Move a pi over four rotation block past its' neighbor block when the blocks anti-commute
 
@@ -257,9 +257,11 @@ class PauliOpCircuit(object):
             raise Exception("First operand must be +-pi/4 Pauli rotation")
 
         if not PauliOpCircuit.are_commuting(self.ops[index], self.ops[next_block]):
-            self.__swap_adjacent_anticommuting_blocks(index)
+            self._swap_adjacent_anticommuting_blocks(index)
+        elif PauliOpCircuit.are_commuting(self.ops[index], self.ops[next_block]):
+            self._swap_adjacent_commuting_blocks(index)
         else:
-            self.__swap_adjacent_commuting_blocks(index)
+            raise Exception("The two blocks must commute or anti-commute!")
 
     def circuit_has_measurements(self) -> bool:
         """Check if circuit has any Measurement blocks.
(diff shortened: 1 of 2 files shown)


src/lsqecc/pauli_rotations/circuit.py
```diff
@@ -191,15 +191,62 @@ def remove_y_operators_from_circuit(self, start_index: int = 0) -> None:
 
             i += 1
 
-    def commute_pi_over_four_rotation(self, index: int) -> None:
-        """Commute a rotation block pass its neighbor block.
+    def __swap_adjacent_commuting_blocks(self, index: int) -> None:
```
gwwatkin: I think were using the convention to use a single underscore for private methods, but it doesn't really matter

src/lsqecc/pauli_rotations/circuit.py
```diff
@@ -191,15 +191,62 @@ def remove_y_operators_from_circuit(self, start_index: int = 0) -> None:
...
+        """
+        Move a pi over four rotation block past its' neighbor block when the blocks commute
+
+        Args:
+            index (int): Index of the targeted block in the current circuit
+        """
```
gwwatkin, line 200: This would be a good place for an are commuting assert

src/lsqecc/pauli_rotations/circuit.py
```diff
@@ -191,15 +191,62 @@ def remove_y_operators_from_circuit(self, start_index: int = 0) -> None:
...
+
+    def __swap_adjacent_anticommuting_blocks(self, index: int) -> None:
+        """
+        Move a pi over four rotation block past its' neighbor block when the blocks anti-commute
 
         Args:
-            index (int): Index of the targeted block in the current circuit.
+            index (int): Index of the targeted block in the current circuit
 
         """
```
gwwatkin, line 216: And here an anti commuting assert

gwwatkin (approved):
:+1: 

Left two notes in places where one could trivially add an assert, now that those blocks of code are separate from the condition to call them and so there is an invariant.

src/lsqecc/pauli_rotations/circuit.py
isolatedinformation: The asserts are handled in the function `swap_adjacent_blocks()`. The respective private functions are only called if they commute or anti-commute

gwwatkin: Yes, they are now, but it's an invariant that needs to hold in the future. It's a strategic thing to protect code when adding new features and refactoring, because a future programmer might not be aware of this invariant (I know it's encoded in the name, but still...).

It just seem like a trivial change that can save so headaches to someone in the future

isolatedinformation: I can add them not an issue! I did not this was a good coding practice :)

src/lsqecc/pauli_rotations/circuit.py
```diff
@@ -209,35 +256,12 @@ def commute_pi_over_four_rotation(self, index: int) -> None:
...
-
-        temp = self.ops[index]
-        self.ops[index] = self.ops[next_block]
-        self.ops[next_block] = temp
-        # print(self.render_ascii())
+            self._swap_adjacent_anticommuting_blocks(index)
+        elif PauliOpCircuit.are_commuting(self.ops[index], self.ops[next_block]):
+            self._swap_adjacent_commuting_blocks(index)
+        else:
+            raise Exception("The two blocks must commute or anti-commute!")
```
alexnguyenn:
```suggestion
        else:
            self._swap_adjacent_commuting_blocks(index)
```
`are_commuting` will only return True or False so this is unnecessary. We will do the error handling in there.

isolatedinformation: Made the suggested changes.

alexnguyenn (approved):

PR #185, opened 2021-12-07
Refactoring `remove_y_operators_from_circuit`

### LINKED ISSUE #184: Refactoring `remove_y_operators_from_circuit`
`remove_y_operators_from_circuit` is a critical component for our compiler. Right now it contains too many moving parts which makes it hard for us to write tests. 

It also uses the `pop` mechanism that we previously decided against. It was meant to move the newly generated pi/4 rotation on the RHS towards the end of the circuit ~~so we don't have to run `apply_transformation` again~~:

https://github.com/latticesurgery-com/lattice-surgery-compiler/blob/942c5c4010daeae4b1d815ae47c4b01770f22cc9/src/lsqecc/pauli_rotations/circuit.py#L174-L180

So I am planning to:
* Divide the whole thing into smaller parts - starting with the Y op removing logic for a single block.
* Add test. Will use examples from Litinski's but might need help with verification. 
* Remove `pop` logic. 

This is opened for discussion. LMK if I am missing anything here.

alexnguyenn:
Closes #184 

Currently still in progress but just want to get the PR out so we can discuss it:
* Move the Y operator removing logic to `PauliRotation`. Looking into circuit.py seems like we are only using this for pi/8 rotations, so I am keeping that. 
* Add some test cases (taken from #35). However we might want to add more (I am wondering whether the current behavior is consistent with -pi/8 rotations).

### DIFF
--- a/src/lsqecc/pauli_rotations/circuit.py
+++ b/src/lsqecc/pauli_rotations/circuit.py
@@ -126,64 +126,19 @@ class PauliOpCircuit(object):
             if circuit_has_measurements:
                 self.ops.pop()
 
-    def remove_y_operators_from_circuit(self, start_index: int = 0) -> None:
-        """Removes Y operators from pi/8 and measurement blocks. To be called
-        after pi/4 rotations have been commuted to the end of the circuit.
-        """
-        i = start_index
-        circuit_has_measurements: bool = self.circuit_has_measurements()
+    def remove_y_operators_from_circuit(self) -> "PauliOpCircuit":
+        """Return a Y-operator-free copy of the current circuit."""
+        y_free_circuit = PauliOpCircuit(self.qubit_num, self.name)
 
-        while i < len(self.ops):
-            pauli_block = self.ops[i]
-            if isinstance(pauli_block, Measurement) or (
-                cast(PauliRotation, pauli_block).rotation_amount
-                in {Fraction(1, 8), Fraction(-1, 8)}
+        for block in self.ops:
+            if isinstance(block, Measurement) or (
+                cast(PauliRotation, block).rotation_amount in {Fraction(1, 8), Fraction(-1, 8)}
             ):
-                y_op_indices = list()
-
-                # Find Y operators and modify them into X operators
-                for j in range(self.qubit_num):
-                    if pauli_block.ops_list[j] == PauliOperator.Y:
-                        y_op_indices.append(j)
-                        pauli_block.ops_list[j] = PauliOperator.X
-
-                if y_op_indices:
-                    right_block_indices = list()
-
-                    # For even numbers of Y operators, add 2 additional pi/4
-                    # rotations (one on each side)
-                    if len(y_op_indices) % 2 == 0:
-                        first_operator = y_op_indices.pop(0)
-                        self.add_single_operator(first_operator, PauliOperator.Z, Fraction(1, 4), i)
-                        self.add_single_operator(
-                            first_operator, PauliOperator.Z, Fraction(-1, 4), i + 2
-                        )
-                        right_block_indices.append(i + 4)
-                        i += 1
-
-                    # Add 2 pi/4 rotations on each side of the Pauli block
-                    new_block = [
-                        PauliOperator.Z if i in y_op_indices else PauliOperator.I
-                        for i in range(self.qubit_num)
-                    ]
-                    self.add_pauli_block(PauliRotation.from_list(new_block, Fraction(1, 4)), i)
-                    self.add_pauli_block(PauliRotation.from_list(new_block, Fraction(-1, 4)), i + 2)
-                    right_block_indices.append(i + 2)
-                    i += 1
-
-                    # Commute the right (new) pi/4 rotations towards the end of the circuit
-                    for right_block_index in right_block_indices:
-                        while right_block_index + 1 < len(self.ops):
-                            self.swap_adjacent_blocks(right_block_index)
-                            right_block_index += 1
-                        if circuit_has_measurements:
-                            self.ops.pop()
+                y_free_circuit.ops.extend(block.get_y_free_equivalent())
             else:
-                # This is assuming pi/4 rotations (not including ones from this operation)
-                # have been commuted to the end of the circuit.
-                break
+                y_free_circuit.add_pauli_block(copy.deepcopy(block))
 
-            i += 1
+        return y_free_circuit
 
     def _swap_adjacent_commuting_blocks(self, index: int) -> None:
         """
--- a/src/lsqecc/pauli_rotations/circuit.py
+++ b/src/lsqecc/pauli_rotations/circuit.py
@@ -126,7 +126,7 @@ class PauliOpCircuit(object):
             if circuit_has_measurements:
                 self.ops.pop()
 
-    def remove_y_operators_from_circuit(self) -> "PauliOpCircuit":
+    def get_y_free_equivalent(self) -> "PauliOpCircuit":
         """Return a Y-operator-free copy of the current circuit."""
         y_free_circuit = PauliOpCircuit(self.qubit_num, self.name)
 
--- a/src/lsqecc/pauli_rotations/rotation.py
+++ b/src/lsqecc/pauli_rotations/rotation.py
@@ -151,6 +151,42 @@ class PauliProductOperation(ABC):
             ]
         )
 
+    @abstractmethod
+    def get_y_free_equivalent(self):
+        """Return the equivalent of current block but without Y operator."""
+        y_op_indices = list()
+        y_free_block = copy.deepcopy(self)
+
+        # Find Y operators and modify them into X operators
+        for index, operator in enumerate(self.ops_list):
+            if operator == PauliOperator.Y:
+                y_op_indices.append(index)
+                y_free_block.ops_list[index] = PauliOperator.X
+
+        if not y_op_indices:
+            return [y_free_block]
+
+        left_rotations = list()
+        right_rotations = list()
+        if len(y_op_indices) % 2 == 0:
+            # For even numbers of Y operators, add 2 additional pi/4 rotations (one on each side)
+            first_y_operator = y_op_indices.pop(0)
+            new_rotation_ops = [
+                PauliOperator.Z if i == first_y_operator else PauliOperator.I
+                for i in range(self.qubit_num)
+            ]
+            left_rotations.append(PauliRotation.from_list(new_rotation_ops, Fraction(1, 4)))
+            right_rotations.append(PauliRotation.from_list(new_rotation_ops, Fraction(-1, 4)))
+
+        new_rotation_ops = [
+            PauliOperator.Z if i in y_op_indices else PauliOperator.I for i in range(self.qubit_num)
+        ]
+
+        left_rotations.append(PauliRotation.from_list(new_rotation_ops, Fraction(1, 4)))
+        right_rotations.append(PauliRotation.from_list(new_rotation_ops, Fraction(-1, 4)))
+
+        return left_rotations + [y_free_block] + right_rotations
+
 
 class PauliRotation(PauliProductOperation, coc.ConditionalOperation):
     """Class for representing a Pauli Product Rotation Block."""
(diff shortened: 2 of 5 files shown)


src/lsqecc/pauli_rotations/rotation.py
```diff
@@ -192,6 +192,43 @@ def from_list(pauli_ops: List[PauliOperator], rotation: Fraction) -> "PauliRotat
...
+        y_op_indices = list()
+
+        # Find Y operators and modify them into X operators
+        for index, operator in enumerate(self.ops_list):
+            if operator == PauliOperator.Y:
+                y_op_indices.append(index)
+                self.ops_list[index] = PauliOperator.X
+
+        if not y_op_indices:
+            return left_rotations, right_rotations
```
isolatedinformation: Why are we returning empty lists here? I understand no operator has been removed, but could we not just `return None`?

src/lsqecc/pauli_rotations/rotation.py
```diff
@@ -192,6 +192,43 @@ def from_list(pauli_ops: List[PauliOperator], rotation: Fraction) -> "PauliRotat
...
+            right_rotations.append(PauliRotation.from_list(new_rotation_ops, Fraction(-1, 4)))
+
+        new_rotation_ops = [
+            PauliOperator.Z if i in y_op_indices else PauliOperator.I for i in range(self.qubit_num)
+        ]
+
+        left_rotations.append(PauliRotation.from_list(new_rotation_ops, Fraction(1, 4)))
+        right_rotations.append(PauliRotation.from_list(new_rotation_ops, Fraction(-1, 4)))
+
+        return left_rotations, right_rotations
```
isolatedinformation: What is the workflow here? Adding the left and right rotations into the circuit will take place in `circuit.py` ?

tests/pauli_rotations/rotation_test.py
```diff
@@ -235,6 +235,31 @@ def test_to_latex(self, input, expected):
         assert input.to_latex() == expected
 
 
+def test_cases_remove_y_operators():
+    # TODO: Add more cases here
```
isolatedinformation: what are the 3 entries in the tuple of each case, (rotations, left_rotations, right_rotations) ?

src/lsqecc/pauli_rotations/rotation.py
alexnguyenn: For me it was to keep it consistent with the typing I declared above (and thus keeping mypy happy).

src/lsqecc/pauli_rotations/rotation.py
alexnguyenn: Yep, basically:
```
# Before
---rotation---

# After
---left_rotations---rotation_without_Y---right_rotations---
```

I will make it clearer in the code and docstring later.

tests/pauli_rotations/rotation_test.py
alexnguyenn: So we have:
* The input rotation
* The input rotation after Y operators removal is performed
* The returned `left_rotations` and `right_rotations`

I also noticed I placed the tests in the wrong place. Fixed that.

src/lsqecc/pauli_rotations/rotation.py
gwwatkin: There is some discussion in this issue https://github.com/latticesurgery-com/lattice-surgery-compiler/issues/35:

From Litinski's GoSC:
image

src/lsqecc/pauli_rotations/rotation.py
```diff
@@ -192,6 +192,43 @@ def from_list(pauli_ops: List[PauliOperator], rotation: Fraction) -> "PauliRotat
             r.change_single_op(i, op)
         return r
 
+    def remove_y_operators(self) -> Tuple[List["PauliRotation"], List["PauliRotation"]]:
```
gwwatkin: This is already a huge improvement in terms of engineering. I can only suggest going even a step further. Just as a good software heuristic, I suggest considering to apply the principle of least privilege and making this method a non modifier. Something like 
```py
def get_y_operator_free_equivalent(self) -> List["PauliRotation"]:
```
Of course usage in the circuit would have to change as well.

src/lsqecc/pauli_rotations/rotation.py
gwwatkin: With my suggestion above the workflow would become to replace the rotations with the Y operator with the list returned by `rotation.get_y_operator_free_equivalent`

gwwatkin (approved):
I'm already fine with how it is now, if you want to go ahead. I trust @isolatedinformation's expertise for the test cases.

However, I suggest taking the refactoring even further, since we are here. Sometimes only after you've spent some time refactoring something, an even cleaner solution becomes visible.

src/lsqecc/pauli_rotations/rotation.py
alexnguyenn: Just updated the code based on the suggestion. Can you have a look?

gwwatkin: Looks good, this is exactly what I meant. I'm assuming you will integrate them in Circuit as well?

alexnguyenn: Did it in the last commit, also made the circuit method to be non-modifier. Might need a name update and update to the codes that are calling it.

tests/pauli_rotations/rotation_test.py
alexnguyenn: I have updated the test to follow @gwwatkin's suggestion, we should have only `input_rotation` and `output` now.

src/lsqecc/pauli_rotations/rotation.py
alexnguyenn: Implemented @gwwatkin's suggestion

src/lsqecc/pauli_rotations/circuit.py
```diff
@@ -126,64 +126,19 @@ def apply_transformation(self, start_index: int = 0) -> None:
...
-                            self.swap_adjacent_blocks(right_block_index)
-                            right_block_index += 1
-                        if circuit_has_measurements:
-                            self.ops.pop()
+                y_free_circuit.ops.extend(block.get_y_free_equivalent())
             else:
-                # This is assuming pi/4 rotations (not including ones from this operation)
-                # have been commuted to the end of the circuit.
-                break
+                y_free_circuit.add_pauli_block(copy.deepcopy(block))
```
gwwatkin, line 139: This is still assuming pi/4 rotations have been commuted already. I think we can actually remove this if statement and apply ` y_free_circuit.ops.extend(block.get_y_free_equivalent())` to all circuit elements

src/lsqecc/pauli_rotations/circuit.py
```diff
@@ -126,64 +126,19 @@ def apply_transformation(self, start_index: int = 0) -> None:
             if circuit_has_measurements:
                 self.ops.pop()
 
-    def remove_y_operators_from_circuit(self, start_index: int = 0) -> None:
-        """Removes Y operators from pi/8 and measurement blocks. To be called
-        after pi/4 rotations have been commuted to the end of the circuit.
-        """
-        i = start_index
-        circuit_has_measurements: bool = self.circuit_has_measurements()
+    def remove_y_operators_from_circuit(self) -> "PauliOpCircuit":
```
gwwatkin: I think this is a  good change. `remove_y_operators_from_circuit` looks pretty clean now. I agree that maybe you want to change the name. 

Will also have to update the pipeline code that calls `remove_y_operators_from_circuit` but that should be trivial.

src/lsqecc/pauli_rotations/rotation.py
```diff
@@ -150,6 +151,42 @@ def get_ops_map(self) -> Dict[int, PauliOperator]:
...
+        right_rotations = list()
+        if len(y_op_indices) % 2 == 0:
+            # For even numbers of Y operators, add 2 additional pi/4 rotations (one on each side)
+            first_y_operator = y_op_indices.pop(0)
+            new_rotation_ops = [
+                PauliOperator.Z if i == first_y_operator else PauliOperator.I
+                for i in range(self.qubit_num)
+            ]
+            left_rotations.append(PauliRotation.from_list(new_rotation_ops, Fraction(1, 4)))
+            right_rotations.append(PauliRotation.from_list(new_rotation_ops, Fraction(-1, 4)))
```
gwwatkin, line 179: Ok taking a more careful look, we've only shown this to be true as long as we are assuming `self` is a pi/8 rotation.  Because adding the corrective pi/4 rotations is just the math from Litinski's Figure 10. However it might also work in other cases

Y operators need to be removed otherwise for pi/4 rotations and measurement. The way the mat was proved was by adding the corrective terms both on one side as `I = exp(-i*P*pi/4) exp(-i*P*pi/4)` and then commuting one of the terms past the original rotation, which made the Y disappear: `exp(-i*P*phi) = exp(-i*P*pi/4) exp(i*P*pi/4) exp(-i*P*phi) = exp(-i*P*pi/4) exp(-i*Q*phi) exp(i*P*pi/4)` (same thing for the case where two corrective rotations per sideare needed). So I would expect  the same calculation to work for `phi=pi/4` and possibly for measurement since it is subject to the same commutation rules. 

@isolatedinformation I't like you input as well, here

src/lsqecc/pauli_rotations/circuit.py
gwwatkin: Ok I see that the code inside `get_y_free_equivalent` actually assumes this is a pi/8 rotation ... but not measurement. Anyhow that assumption needs to be checked, see my other comment.

What I think will happen in the end is that we'll be able to remove this if statement for a unified solution

gwwatkin (commented):
This is becoming a subtle issue about consolidating the behavior of the compiler. Things I think need to be addressed.
- [x] Check the correctness of how we remove Y operators from pi/4 rotations
- [x] Check the correctness of how we remove Y operators from measurements
- [x] Find out if we can remove the branch on pi/8 rotations (we can)
- [x] Adapt usages of `remove_y_operators_from_circuit` and rename
- [ ] Tests for (separate pull request?)

src/lsqecc/pauli_rotations/circuit.py
alexnguyenn: Yeah I have made the method available for measurements and pi/8 rotations as it was how we were doing in the original implementation. The `if` statement there is based on that too. 

But yeah we would like to have more proofs before merging this in.

src/lsqecc/pauli_rotations/rotation.py
isolatedinformation: Litinski's case is true for `phi=pi/8`. Consider `exp(-i*P*phi) = cos(phi) I - i sin(phi) P`. For `phi=pi/8`, `cos(phi)!=sin(phi)`. So we can effectively rule it out it is just true for that particular angle as it they are not the same

gwwatkin: WLOG phi=pi/8 then.

gwwatkin: It remains to show this for measurements

src/lsqecc/pauli_rotations/circuit.py
gwwatkin: Ok based on our [other conversation](https://github.com/latticesurgery-com/lattice-surgery-compiler/pull/185/files/a80dc0ff52646d8595ada1f0c409b64698646de1#r767273177) I think we have enough evidence group this special case with pi/8. There is no difference between angles.

src/lsqecc/pauli_rotations/rotation.py
alexnguyenn: @isolatedinformation can you check if it is the same for measurements?

isolatedinformation: I have written the proofs for both general angle Y rotations and Measurements in the proofs goodnotes notebook. It stands true as conjectured.

gwwatkin:
As for removing Y operators from measurements, meeting with @isolatedinformation it cam up that:
 - It is possible to do it
 - We don't need to for multi body measurements because if the pauli rotations have already been purged of Ys, all the multi body measurements will be Y operator free, **as long as the Litinski Transform isn't running**, since it can introduce Y operators in measurements.
 - The only way we get Y measurements is if the user explicitly requires them, for now we will just ignore that feature

src/lsqecc/pauli_rotations/rotation.py
```diff
@@ -192,6 +229,15 @@ def from_list(pauli_ops: List[PauliOperator], rotation: Fraction) -> "PauliRotat
             r.change_single_op(i, op)
         return r
 
+    def get_y_free_equivalent(self) -> List["PauliRotation"]:
+        """Return the equivalent of current Pauli Rotation but without Y operator.
+        Only supports pi/8 rotations at the moment.
+        """
+        if self.rotation_amount not in (Fraction(1, 8), Fraction(-1, 8)):
```
isolatedinformation: This check is irrelevant because the remove operation stands true for all angles. and any Y operator can be removed

tests/pauli_rotations/rotation_test.py
```diff
@@ -271,6 +381,21 @@ def test_pauli_product_op_init(self):
...
+    def test_remove_y_operators_no_y_op(self):
+        r = PauliRotation.from_list([X, Z, I, Z], Fraction(1, 8))
+        assert r.get_y_free_equivalent() == ([], [])
+        assert r.ops_list == [X, Z, I, Z]
+
+    @pytest.mark.parametrize("rotation, new_rotation, lhs, rhs", test_cases_remove_y_operators())
+    def test_remove_y_operators_1(self, rotation, new_rotation, lhs, rhs):
+        assert rotation.get_y_free_equivalent() == (lhs, rhs)
+        assert rotation == new_rotation
+
```
isolatedinformation: @alexnguyenn are these tests still required? 1. They are not passing; 2. Tests have been included seperately for PauliRotation and Measurement in their respective classes

gwwatkin: Can safely remove I think

gwwatkin: ops just saw that the question was @alexnguyenn

alexnguyenn: sr I am OOTL - why are those failing now?

isolatedinformation: It is failing because how the arguments to the test are passed. That can be fixed, but my question is if do they need to be? Previously, line 396 was 
```
assert rotation.remove_y_operator() == (lhs, rhs)
```
I thought changing it to `get_y_free_equuvalent` would fix it, But it fails because of incompatible assert statements. I now augmented the tests for both  `PauliRotations` and `Measurements` for removing y operators. Since they are already checked in the tests above, can these test be removed ?(line 384-398)

alexnguyenn: Ok from what I can see I already removed them in https://github.com/latticesurgery-com/lattice-surgery-compiler/pull/185/commits/fcf36f7be6afa58acc01cefb32b3d89cb52fa9ac above. I am confused to why are they added back in? 

Was this caused by the merge commit you made above?

isolatedinformation: Oh it could be. I did update my local before pushing. But I shall remove it now

alexnguyenn: I think the problem right now is I don't know what did the merge commit change here - we might even have to revert.

isolatedinformation: we can revert to last known checkpoint and add the changes I made in a new PR. Does that sound okay? we can revert to a80dc0f and I can add the changes I made from e407c8a to that

isolatedinformation (approved):

gwwatkin (approved):
Looks good. There still are some untested lines, but they are relating to the circuit wrapper for `get_t_free_equivalent`. It can merge how it is.

gwwatkin:
As @alexnguyenn pointed out we were forgetting to update the calls to the modified function

isolatedinformation (approved):

PR #191, opened 2021-12-19
Convert opflow imports from aqua to qiskit core

### LINKED ISSUE #187: [Feature Request] Migrate Qiskit Aqua
Issue Description
-----------------

From [qiskit-aqua's readme](https://github.com/Qiskit/qiskit-aqua):
> As of version 0.9.0, released on 2nd April 2021, Qiskit Aqua has been deprecated with its support ending and eventual archival being no sooner than 3 months from that date. The function provided by Qiskit Aqua is not going away rather it has being split out to separate application repositories, with core algorithm and operator function moving to qiskit-terra. Please see the Migration Guide section below for more detail. We encourage you to migrate over at your earliest convenience.

Proposed Solution
-----------------
I think the only feature we are using from aqua is operatorflow. According to the [migration guide](https://github.com/Qiskit/qiskit-aqua#migration-guide), this is just being moved to terra, which will still be part of qiskit, so it should just be a matter of changing imports.

The time frame for this migration is 3 months, so end of February. It could be that by then we won't be using operatorflow for simulation any longer, in which case there won't be anything to do.

gwwatkin:
As outlined in #187 qiskit is deprecating aqua which we use for opflow. Since there are no other occurrences of `aqua` in our project sources, just changing the imports.

Closes #187.

### DIFF
--- a/debug/debug_state_separator.py
+++ b/debug/debug_state_separator.py
@@ -9,7 +9,7 @@ from qiskit_opflow_utils import StateSeparator
 import webgui.lattice_view
 from fractions import Fraction
 
-import qiskit.opflow as qk
+import qiskit.opflow as qkop
 import qiskit.quantum_info as qkinfo
 
 import math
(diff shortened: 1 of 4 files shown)


debug/debug_state_separator.py
```diff
@@ -9,7 +9,7 @@
 import webgui.lattice_view
 from fractions import Fraction
 
-import qiskit.aqua.operators as qk
+import qiskit.opflow as qk
```
alexnguyenn: We are both using `qiskit.opflow.*` and `qk.*` in our code - can you unify them under 1 format? I am thinking `qkop` or something similar.

alexnguyenn (approved):

PR #192, opened 2021-12-19
Point to `dev` contributing guide

gwwatkin:
Makes sense even when `dev` is not going to be our main branch anymore (which we can start thinking about), because we want to point people to the most up to date information.


alexnguyenn (approved):
Good spot

Issue #138, opened 2021-11-08
[Feature Request]  Speed up simulation by removing patches that are not on the lattice from the state vector 


gwwatkin:
Issue Description
-----------------
Simulation is a critical bottleneck right now and it is because of simulating useless patches.

Right now simulation of longer circuits happens at a much slower speed because every patch that is ever on the lattice contributes to one big state vector. In particular state vector size grows exponentially with y-depth and t-depth, because each of these operations is associated with an new patch that contributes an extra qubit in the logical lattice computation, doubling it's size.

Even these patches aren't contributing to the computation for nearly the whole length of the computation. This makes simulating circuits of just a couple lines infeasible without y operator removal.

Addressing this issue will greatly expand the set of circuits we can test.

Proposed Solution
-----------------

Here are some possible solutions:
- Exclude patches not on the lattice from the state vector
  - Might be the easiest to implement 
- Break the global state into smaller vectors based on separability
   - We already have a separability algorithm
   - Can cache what was not entangled from previous operations  
   - Speeds up all not entangled computations (E.g. different modules running in parallel)
- Leverage an existing simulator to simulate one operation at a time

alexnguyenn:
@isolatedinformation I think your opinion would be helpful here

isolatedinformation:
- **Excluding Patches**
If I understand what you mean correctly, I think we would run into problems similar to `ops.pop()`. If only the correctness of the smaller statevector, this would be alright I think. 

- **Seperability and Caching**
    1. Do you mean, there is already a separability algorithm implemented in the package, or do we know which algorithm to implement?
    2. I like the idea of caching better. Because we are not really getting rid of the state vector. Just caring about the parts we need. 
    3. The only disadvantage I see, if it turns out the state vector is not seperable we would not get any speedup
    4. This would also be more scalable, imo. 
      
- **External Simulators** 
This choice depends more on how we want to depend on existing simulators. I remember @alexnguyenn he wants to reduce the dependency on Qiskit. Relying on other simulators would prevent us from reinventing the wheel again.

PR #197, opened 2021-12-26
StateSeparator Tests

gwwatkin:
- [x] `trace_dict_state`
 - [x] `trace_to_density_op`
 - [x] `separate`
 - [x] `get_separable_qubits`

Test required to work on the functionality for #138


src/lsqecc/simulation/qiskit_opflow_utils.py
```diff
@@ -85,5 +84,15 @@ def get_separable_qubits(dict_state: qkop.DictStateFn) -> Dict[int, qkop.DictSta
...
+
+
+def to_dict_fn(vector_state: qkop.OperatorBase) -> qkop.DictStateFn:
+    # Switch two identical cases to keep mypy happy
+    if isinstance(vector_state, qkop.VectorStateFn):
+        return cast(qkop.DictStateFn, vector_state.to_dict_fn())
+    elif isinstance(vector_state, qkop.SparseVectorStateFn):
+        return cast(qkop.DictStateFn, vector_state.to_dict_fn())
+    else:
+        raise NotImplementedError(f"Conversion to_dict_fn of {repr(type(vector_state))}")
```
alexnguyenn, line 98: codecov is asking for a test case here 😅 (L98).

alexnguyenn (approved):
Nice to see more tests added. I saw you did run into some troubles with mypy - personally I don't think typing is that necessary for tests? 

Can't comment on the correctness so we need @isolatedinformation's comments.

gwwatkin:
@alexnguyenn these tests juggle a lot of similar but not the same types like `StateVector`, vector `StateFn`s and numpy arrays so I ended up writing annotations to make sense of what was going on, but generally I would say typing isn't necessary for tests

gwwatkin:
@isolatedinformation  FYI

PR #199, opened 2021-12-28
ProjectiveMeasurement tests

gwwatkin:
Tests required to work on the functionality for #138

PR #209, opened 2022-01-13
Qubit Mapper tests

gwwatkin:
Testing the mechanism that associates patches to qubits. These tests are not the depest because how this is tightly tangled with logical computation. For now I'm happy with this because we have a vast agenda of other things to attend to and there will be the simulation integration tests to give us more confidence that these mechanisms are working

WIP because #199 needs to go in first, then the diff will be cleaned of the protective measurement tests

Only the last 3 commits are relevant


gwwatkin:
closing in favor of #211

PR #211, opened 2022-01-17
Qubit Mapper Tests

gwwatkin:
Testing the mechanism that associates patches to qubits. These tests are not the depest because how this is tightly tangled with logical computation. For now I'm happy with this because we have a vast agenda of other things to attend to and there will be the simulation integration tests to give us more confidence that these mechanisms are working

### DIFF
--- a/tests/simulation/logical_patch_state_simulation_test.py
+++ b/tests/simulation/logical_patch_state_simulation_test.py
@@ -221,4 +221,46 @@ class TestPatchToQubitMapper:
         assert len(set(patch_ids)) == 3  # check that they are all different
 
     def test___init__(self):
-        pass
+        c = circuit.PauliOpCircuit(2)
+        c.add_pauli_block(circuit.PauliRotation.from_list([X, I], Fraction(1, 2)))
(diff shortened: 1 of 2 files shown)


tests/simulation/logical_patch_state_simulation_test.py
```diff
@@ -163,3 +168,99 @@ def test_pauli_product_measurement_distribution(
...
+    [
+        [("a", 0.50), ("b", 0.30), ("c", 0.20)],
+        [
+            ("a", 1.0),
+            ("b", 0),
+            ("c", 0),
+        ],
+    ],
+)
+def test_proportional_choice(assoc_data_prob):
```
isolatedinformation, line 184: what does this function do? The tests look good, but I would like what this function is trying to achieve

isolatedinformation (commented):
Looks good. just let me know what 'proportional_choice()' does.

isolatedinformation (approved):

tests/simulation/logical_patch_state_simulation_test.py
gwwatkin: Internal use. Noted in comment

PR #213, opened 2022-01-18
Add LazyTensorOp

gwwatkin:
Needed by #138. Allows us to keep track of the global state of the patches on the lattice keeping different state-vectors for parts that are not entangled, like ancilla patches waiting to be used or that are already measured.

Has an API to apply operators to this state without worrying about in which state vector the patch is.

First round of functionality and tests. Later will add ways to entangle patches in and use it `logical_patch_state_simulation`

### DIFF
--- a/tests/simulation/lazy_tensor_op_test.py
+++ b/tests/simulation/lazy_tensor_op_test.py
@@ -32,7 +32,7 @@ from tests.simulation.numpy_matrix_assertions import (
 
 class TestLazyTensorOp:
     @pytest.mark.parametrize(
-        "first, second, expected",
+        "rhs, lhs, expected",
         [
             ([qkop.Zero], [qkop.X], [qkop.One]),
             ([qkop.Plus], [qkop.Z], [qkop.Minus]),
@@ -42,12 +42,12 @@ class TestLazyTensorOp:
     )
     def test_apply_matching_tensors(
         self,
-        first: List[qkop.OperatorBase],
-        second: List[qkop.OperatorBase],
+        rhs: List[qkop.OperatorBase],
+        lhs: List[qkop.OperatorBase],
         expected: List[qkop.OperatorBase],
     ):
-        result: LazyTensorOp[qkop.OperatorBase] = LazyTensorOp(first).apply_matching_tensors(
-            LazyTensorOp(second)
+        result: LazyTensorOp[qkop.OperatorBase] = LazyTensorOp(rhs).apply_matching_tensors(
+            LazyTensorOp(lhs)
         )
         assert len(result.ops) == len(expected)
         for i, op in enumerate(expected):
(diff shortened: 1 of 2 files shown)


tests/simulation/lazy_tensor_op_test.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+class TestLazyTensorOp:
+    @pytest.mark.parametrize(
+        "first, second, expected",
+        [
+            ([qkop.Zero], [qkop.X], [qkop.One]),
+            ([qkop.Plus], [qkop.Z], [qkop.Minus]),
+            ([qkop.X], [qkop.Z], [1j * qkop.Y]),
+            ([qkop.X, qkop.I ^ qkop.Z], [qkop.I, qkop.Y ^ qkop.I], [qkop.X, qkop.Y ^ qkop.Z]),
```
isolatedinformation, line 40: is it called Lazy because the Hilbert space which the tensor applies is not explicitly specified? for example `[qkop.X, qkop.I ^ qkop.Z]` does not explicity specify which of two qubits X applies to ?

tests/simulation/lazy_tensor_op_test.py
```diff
@@ -0,0 +1,174 @@
...
+        "qubit_idx, "
+        "expected_index_of_tensor_operand, "
+        "expected_index_within_tensor_operand ",
+        [
+            ([qkop.I], 0, 0, 0),
+            ([qkop.I ^ qkop.I], 0, 0, 0),
+            ([qkop.I ^ qkop.I], 1, 0, 1),
+            ([qkop.I, qkop.I], 1, 1, 0),
+            ([qkop.I ^ qkop.I, qkop.I], 1, 0, 1),
+            ([qkop.I ^ qkop.I, qkop.I], 1, 0, 1),
```
isolatedinformation, line 152: Line 151 and 152 are redundant

isolatedinformation:
Looks good! I have a feeling the indexing might get confusing. for example
```
            (
                [qkop.I, qkop.Y ^ qkop.I, qkop.I ^ qkop.X, qkop.H],
                [qkop.Z, qkop.I ^ qkop.X, qkop.X ^ qkop.Z, qkop.I],
                [qkop.Z, qkop.Y ^ qkop.X, qkop.X ^ (qkop.Y * 1j), qkop.H],
            )
```
maybe some documentation on that might help. it takes a while to figure out the exact order of application and not everyone might take a look at the tests. I approved it. The rest seems solid, but the ordering might be a point of possible confusion and we should have some examples in the documentation imo

isolatedinformation (approved):

tests/simulation/lazy_tensor_op_test.py
gwwatkin: As discussed over zoom, it's lazy in the sense that it is lazy evaluated. Added comment in code

tests/simulation/lazy_tensor_op_test.py
gwwatkin: Nice catch

gwwatkin:
Clarified the left handedness of the application of LazyTensorOps in the class. As for the tests I just gave better names (`rhs` and `lhs`) to the arguments, but the parametrize lists are still in a counterintuitive order, but I think we can live with that as these are just unit tests for an internal tool

gwwatkin:
Some changes were missed due to me being silly with pre-commit. Added in #214

PR #217, opened 2022-01-21
Test for `LogicalLatticeComputation`

isolatedinformation:
- tests for _load_circuit() yet to written. Will be written in conjunction with the tests for RotationsComposer
- currently failing mypy checks with error
``` 
tests/logical_lattice_ops/logical_lattice_ops_test.py:143: error: Argument 1 to "circuit_to_patch_measurement" of "LogicalLatticeComputation" has incompatible type "PauliProductOperation"; expected "Measurement"
tests/logical_lattice_ops/logical_lattice_ops_test.py:155: error: Argument 1 to "circuit_to_patch_measurement" of "LogicalLatticeComputation" has incompatible type "PauliProductOperation"; expected "Measurement"
```

### DIFF
--- a/src/lsqecc/logical_lattice_ops/logical_lattice_ops.py
+++ b/src/lsqecc/logical_lattice_ops/logical_lattice_ops.py
@@ -118,8 +118,10 @@ class LogicalLatticeComputation:
                 self.ops.append(current_op)
 
     def circuit_to_patch_measurement(
-        self, m: Measurement
+        self, m: PauliProductOperation
     ) -> Union[SinglePatchMeasurement, MultiBodyMeasurement]:
+        if not isinstance(m, Measurement):
+            raise TypeError("Make sure the passed argument is of type Measurement")
 
         ret: Dict[uuid.UUID, PauliOperator] = dict()
         for qubit_idx in range(m.qubit_num):
--- a/tests/logical_lattice_ops/logical_lattice_ops_test.py
+++ b/tests/logical_lattice_ops/logical_lattice_ops_test.py
@@ -88,9 +88,7 @@ def generate_tests_count_magic_states():
     with open("assets/demo_circuits/nontrivial_state.qasm") as input_file:
         c1 = segmented_qasm_parser.parse_str(input_file.read())
 
-    with open(
-        "/home/varunseshadri/projects/lattice-surgery-compiler/assets/demo_circuits/bell_pair.qasm"
-    ) as input_file:
+    with open("assets/demo_circuits/bell_pair.qasm") as input_file:
         c2 = segmented_qasm_parser.parse_str(input_file.read())
 
     return [(c1, 1), (c2, 0)]
@@ -157,6 +155,13 @@ class TestLogicalLatticeComputation:
         assert list(patch_measurement.patch_pauli_operator_map.values()) == list(
             measurement.get_ops_map().values()
         )
+
+    def test_circuit_to_patch_measurement_type_error(self):
+        circuit = PauliOpCircuit(1)
+        circuit.add_pauli_block(PauliRotation.from_list([X], Fraction(1, 4)))
+        logical_comp = LogicalLatticeComputation(circuit)
+        with pytest.raises(TypeError):
+            logical_comp.circuit_to_patch_measurement(logical_comp.circuit.ops[0])
 
     def test__load_circuit(self):
         # TODO: Determine example circuits, dependent on RotationsComposer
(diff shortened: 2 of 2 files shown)


isolatedinformation:
- mypy errors resolved by raising type error in `circuit_to_patch_measurement`. changes in commit `6a84811`

src/lsqecc/logical_lattice_ops/logical_lattice_ops.py
```diff
@@ -118,8 +118,10 @@ def to_lattice_operation(
                 self.ops.append(current_op)
 
     def circuit_to_patch_measurement(
-        self, m: Measurement
+        self, m: PauliProductOperation
     ) -> Union[SinglePatchMeasurement, MultiBodyMeasurement]:
+        if not isinstance(m, Measurement):
+            raise TypeError("Make sure the passed argument is of type Measurement")
```
gwwatkin, line 124: was `mypy` failing to enforce this?

tests/logical_lattice_ops/logical_lattice_ops_test.py
```diff
@@ -63,3 +67,102 @@ def test_get_operating_patches(self):
...
+
+    def test_circuit_to_patch_measurement_type_error(self):
+        circuit = PauliOpCircuit(1)
+        circuit.add_pauli_block(PauliRotation.from_list([X], Fraction(1, 4)))
+        logical_comp = LogicalLatticeComputation(circuit)
+        with pytest.raises(TypeError):
+            logical_comp.circuit_to_patch_measurement(logical_comp.circuit.ops[0])
+
+    def test__load_circuit(self):
+        # TODO: Determine example circuits, dependent on RotationsComposer
```
gwwatkin, line 167: I'm fine if you want to add these in a separate PR

tests/logical_lattice_ops/logical_lattice_ops_test.py
```diff
@@ -63,3 +67,102 @@ def test_get_operating_patches(self):
...
+        "circuit, measurement", generate_tests_circuit_to_multi_body_measurements()
+    )
+    def test_circuit_to_multi_body_measurement(
+        self, circuit: PauliOpCircuit, measurement: Measurement
+    ):
+        logical_comp = LogicalLatticeComputation(circuit)
+        patch_measurement = logical_comp.circuit_to_patch_measurement(logical_comp.circuit.ops[0])
+        assert isinstance(patch_measurement, MultiBodyMeasurement)
+        assert list(patch_measurement.patch_pauli_operator_map.values()) == list(
+            measurement.get_ops_map().values()
```
gwwatkin, line 156: I'm surprised the order works here

gwwatkin (approved):

src/lsqecc/logical_lattice_ops/logical_lattice_ops.py
isolatedinformation: tests/logical_lattice_ops/logical_lattice_ops_test.py:143: error: Argument 1 to "circuit_to_patch_measurement" of "LogicalLatticeComputation" has incompatible type "PauliProductOperation"; expected "Measurement".

mypy was raising this exception. The argument was type-specified to Measurement. However, the ops in a PauliOpCircuit are of the the supertype PauliProductOperation. So I changed the type specification to PauliProductoperation and internally added a type check.

tests/logical_lattice_ops/logical_lattice_ops_test.py
isolatedinformation: yes, I want these to be a part of the RotationsComposer tests PR. they are closely related to each other

tests/logical_lattice_ops/logical_lattice_ops_test.py
isolatedinformation: i think it's because both use the same logic. `get_ops_map()` populates as long as it is not identity. So does `circuit_to_patch_measurement` while populating the `patch_pauli_operator_map()`. Could this a be possible point of failure in the future? I think I wrote tests that are robust. Do you have additional test cases in mind for this particular test?

src/lsqecc/logical_lattice_ops/logical_lattice_ops.py
gwwatkin: you should be able to use `cast` to work around it

tests/logical_lattice_ops/logical_lattice_ops_test.py
gwwatkin: I don't know ... usually when comparing maps you check the keys as sets and then compare key vale pairs or do some kind of sorting. But since this is a test we'll fix it only if it breaks

src/lsqecc/logical_lattice_ops/logical_lattice_ops.py
isolatedinformation: Will keep in mind next time. Should I rewrite this bit now ?

alexnguyenn (approved):

PR #219, opened 2022-01-23
Lazy Tensor Simulation

### LINKED ISSUE #138: [Feature Request]  Speed up simulation by removing patches that are not on the lattice from the state vector

gwwatkin:
If we have a state like:

`|0> otime |1> otimes (|00>+|11>)/sqrt(2) otimes |0>`

We don't need to expanded out to a state vector with 32 entries to apply, say an X gate to the second qubit.

This PR introduces simulation that reduces the size of the state vectors involved in simulation by tracking the list of tensors operands that make up the tensor state. The class used to hold these tensored operands states is `LazyTensorOps` introduced in #213.

Operations supported:
 - [x] Single qubit gates
 - [x] Single qubit measurements
 - [x] Multibody measurements

Closes #138

### DIFF
--- a/src/lsqecc/pauli_rotations/circuit.py
+++ b/src/lsqecc/pauli_rotations/circuit.py
@@ -322,6 +322,14 @@ class PauliOpCircuit(object):
         return ret_circ
 
     @staticmethod
+    def from_list(pauli_op_list: List[PauliProductOperation]):
+        c = PauliOpCircuit(pauli_op_list[0].qubit_num)
+        for op in pauli_op_list:
+            assert c.qubit_num == op.qubit_num
+            c.add_pauli_block(op)
+        return c
+
+    @staticmethod
     def load_reversible_from_qasm_string(quasm_string: str) -> "PauliOpCircuit":
         """Load a string as if it were a QASM circuit. Only supports reversible circuits."""
 
--- a/src/lsqecc/simulation/lazy_tensor_op.py
+++ b/src/lsqecc/simulation/lazy_tensor_op.py
@@ -90,12 +90,21 @@ class LazyTensorOp(Generic[T]):
                 return i, qubit_idx - sums[i]
         raise IndexError(f"Requesting qubit index: {qubit_idx} out of range: {sums[-1]}")
 
-    def merge_operand_with_the_next(self, target_operand_idx: int):
+    def swap_operands(self, operand_idx1: int, operand_idx2: int) -> None:
+        tmp = self.ops[operand_idx2]
+        self.ops[operand_idx2] = self.ops[operand_idx1]
+        self.ops[operand_idx1] = tmp
+
+    def merge_operand_with_the_next(self, target_operand_idx: int) -> None:
         new_ops = [op for i, op in enumerate(self.ops) if i != target_operand_idx + 1]
         new_ops[target_operand_idx] = (
             self.ops[target_operand_idx] ^ self.ops[target_operand_idx + 1]
         )
-        return LazyTensorOp(new_ops)
+        self.ops = new_ops
+
+    def merge_the_first_n_operands(self, n: int) -> None:
+        for i in range(n):
+            self.merge_operand_with_the_next(0)
 
     def separate_last_qubit_of_operand(self, operand_idx: int) -> bool:
         """Returns true if the qubit was not entangled and hence the operation possible"""
@@ -139,7 +148,8 @@ class LazyTensorOp(Generic[T]):
 
     def matching_approx_eq_vector(self, other: "LazyTensorOp[S]", atol: float = 10 ** (-8)) -> bool:
         # TODO check is vector
-        assert self.matches(other)
+        if not self.matches(other):
+            raise LazyTensorOpsNotMatchingException()
         return all(
             [
                 np.allclose(qkutil.to_vector(op1), qkutil.to_vector(op2), atol=atol)
--- a/src/lsqecc/simulation/lazy_tensor_op.py
+++ b/src/lsqecc/simulation/lazy_tensor_op.py
@@ -18,9 +18,10 @@
 
 from typing import Generic, List, Tuple, TypeVar
 
+import numpy as np
 import qiskit.opflow as qkop
 
-from lsqecc.simulation.qiskit_opflow_utils import StateSeparator
+import lsqecc.simulation.qiskit_opflow_utils as qkutil
 
 
 class LazyTensorOpsNotMatchingException(Exception):
@@ -100,10 +101,15 @@ class LazyTensorOp(Generic[T]):
         """Returns true if the qubit was not entangled and hence the operation possible"""
         operand = self.ops[operand_idx]
         assert isinstance(operand, qkop.DictStateFn)
-        maybe_new_operand_from_qubit = StateSeparator.separate(operand.num_qubits - 1, operand)
+        if operand.num_qubits == 1:
+            return True
+
+        maybe_new_operand_from_qubit = qkutil.StateSeparator.separate(
+            operand.num_qubits - 1, operand
+        )
         if maybe_new_operand_from_qubit is None:
             return False
-        self.ops[operand_idx] = StateSeparator.trace_dict_state(
+        self.ops[operand_idx] = qkutil.StateSeparator.trace_dict_state(
             operand, list(range(operand.num_qubits - 1))
         )
         self.ops.insert(operand_idx + 1, maybe_new_operand_from_qubit)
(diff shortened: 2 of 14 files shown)


gwwatkin:
It seems like there is some subtlety about qubit indices appearing as I write tests ... unfortunately will have to address that in this PR because of the large amount of tests of changes :sweat:

gwwatkin:
The tests aren't as fine grained as uni tests should be, as each method's test require other classes to test, but I'm fine with this situation as I think it still gives us good compromise between time spent writing tests and precision of the tests

gwwatkin:
@isolatedinformation  and @alexnguyenn Ready for a first round of review and comments ... unfortunately the bookkeeping for multi-body measurements got out of hand so this PR is far from clean. I'm happy to go over it in a meeting as well

src/lsqecc/pauli_rotations/circuit.py
```diff
@@ -321,6 +321,14 @@ def load_from_pyzx(circuit) -> "PauliOpCircuit":
         print("Gate Missed: ", gate_missed)
         return ret_circ
 
+    @staticmethod
+    def from_list(pauli_op_list: List[PauliProductOperation]):
+        c = PauliOpCircuit(pauli_op_list[0].qubit_num)
+        for op in pauli_op_list:
+            assert c.qubit_num == op.qubit_num
```
isolatedinformation, line 328: Minor Suggestion: I think we should call it `num_qubits` instead `qubit_num`, because later might thought of as returning the index. There is scope for ambiguity. I see it's called `num_qubits` in line 177 in `lazy_tensor_op.py`

gwwatkin: yep we definitely should ... I think this PR is already big enough though

src/lsqecc/simulation/logical_patch_state_simulation.py
```diff
@@ -263,3 +292,184 @@ def apply_logical_operation(self, logical_op: llops.LogicalLatticeOperation):
...
+        ]
+
+        involved_operand_idxs: List[int] = list(
+            set(
+                [
+                    self.logical_state.get_idxs_of_qubit(qubit_idx)[0]
+                    for qubit_idx in involved_qubit_idxs
+                ]
+            )
+        )
```
alexnguyenn: I am confused here - `list` to `set` to `list`?

src/lsqecc/simulation/lazy_tensor_op.py
```diff
@@ -38,7 +39,7 @@ class LazyTensorOp(Generic[T]):
     operands. Has methods to apply operators so that the tensor structure is preserved."""
 
     def __init__(self, ops: List[T]):
-        self.ops = ops
+        self.ops = [op for op in ops]  # TODO rename to operands
```
alexnguyenn:
```suggestion
        self.ops = ops[:]  # TODO rename to operands
```
I think something like this also works as well.

src/lsqecc/simulation/logical_patch_state_simulation.py
```diff
@@ -263,3 +292,184 @@ def apply_logical_operation(self, logical_op: llops.LogicalLatticeOperation):
...
+            )
+            self.logical_state.ops[operand_idx] = cast(qkop.StateFn, symbolic_state.eval())
+
+        elif isinstance(logical_op, llops.MultiBodyMeasurement):
+            # Prepare the state by moving all involved operands to the front
+            self.align_state_with_pauli_op_to_first_operand(logical_op)
+            size_of_first_operand = self.logical_state.ops[0].num_qubits
+
+            # Make a global observable for the operator
+            global_observable = qkop.I ^ size_of_first_operand
```
isolatedinformation: After discussion with @gwwatkin, for Lazy Tensor this means `first_operand_observable`, with the first_operand effectively causing the multibody measurement . `global_observable` might be misleading

src/lsqecc/simulation/lazy_tensor_op.py
```diff
@@ -77,23 +78,111 @@ def apply_more_granular_lazy_tensor(
...
+            f"Requesting qubit index: {qubit_idx} out of range: {operand_boundaries[-1]}"
+        )
+
+    def get_idx_of_first_qubit_in_operand(self, operand_idx: int):
+        return self.get_idx_of_first_qubit_for_each_operand()[operand_idx]
+
+    def swap_operands(self, operand_idx1: int, operand_idx2: int) -> None:
+        tmp = self.ops[operand_idx2]
+        self.ops[operand_idx2] = self.ops[operand_idx1]
+        self.ops[operand_idx1] = tmp
```
alexnguyenn:
```suggestion
        self.ops[operand_idx2], self.ops[operand_idx1] = self.ops[operand_idx1], self.ops[operand_idx2]
```
You can do something like this in Python

src/lsqecc/simulation/logical_patch_state_simulation.py
```diff
@@ -263,3 +292,184 @@ def apply_logical_operation(self, logical_op: llops.LogicalLatticeOperation):
...
+            get_init_state(quuid) for idx, quuid in self.mapper.enumerate_patches_by_index()
+        ]
+        return LazyTensorOp(all_init_states)
+
+    def apply_logical_operation(self, logical_op: llops.LogicalLatticeOperation):
+        """Update the logical state"""
+
+        if not logical_op.does_evaluate():
+            raise Exception(
+                "apply_logical_operation called with non evaluating operation :" + repr(logical_op)
```
alexnguyenn:
```suggestion
                f"apply_logical_operation called with non evaluating operation: {repr(logical_op)}"
```

alexnguyenn (commented):
I put some comments but I think we will need a meeting to go over all of the logic change.

src/lsqecc/simulation/logical_patch_state_simulation.py
gwwatkin: yes to remove duplicates

gwwatkin: Though it looks like it's no longer necessary

alexnguyenn:
pre-commit.ci autofix

gwwatkin:
> I put some comments but I think we will need a meeting to go over all of the logic change.


@alexnguyenn Yes we probably do. Went over the changes with @isolatedinformation, but we can go over them again

src/lsqecc/simulation/logical_patch_state_simulation.py
```diff
@@ -184,14 +186,129 @@ def _get_all_operating_patches(
...
+    def make_simulator(
+        simulator_type: SimulatorType, logical_computation: llops.LogicalLatticeComputation
+    ):
+        if simulator_type == SimulatorType.LAZY_TENSOR:
+            return LazyTensorPatchSimulator(logical_computation)
+        else:
+            return FullStateVectorPatchSimulator(logical_computation)
+
+
+class FullStateVectorPatchSimulator(PatchSimulator):
```
gwwatkin, line 213: This is just adding back the existing state vector simulator, unfortnuately it didn't occur to me that we would probably want to keep both in the end

src/lsqecc/simulation/logical_patch_state_simulation.py
```diff
@@ -307,16 +424,52 @@ def bring_active_operands_to_front(self, logical_op: llops.MultiBodyMeasurement)
...
+                operand_uuid_list[operand_swap_target_counter],
+            )
+            self.mapper.patch_location_to_logical_idx = dict(
+                [
+                    (patch_uuid, idx)
+                    for idx, patch_uuid in enumerate(itertools.chain(*operand_uuid_list))
+                ]
+            )
+
+            operand_swap_target_counter += 1
```
gwwatkin, line 449: Looks like I didn't make a separate commit for this change or lost it in rebase, but all this is doing is fixing a mismatch of indices when swapping operators

src/lsqecc/simulation/logical_patch_state_simulation.py
gwwatkin: Applied suggestion

isolatedinformation (approved):

PR #220, opened 2022-01-25
Separate compilation text

Keelando:
separated compilation text string into a JSON formatted string

compilation text -> input circuit, circuit after pauli rotations, circuit after litinski rotations.

will allow cleaner logic on the web UI

tested

### DIFF
--- a/src/lsqecc/pipeline/json_pipeline.py
+++ b/src/lsqecc/pipeline/json_pipeline.py
@@ -20,7 +20,7 @@ import json
 
 import lsqecc.lattice_array.visual_array_cell as vac
 
-from .lattice_surgery_compilation_pipeline import compile_str
+from .lattice_surgery_compilation_pipeline import Compilation
 
 
 class JsonResponse:
@@ -77,7 +77,7 @@ def handle(json_request: str) -> JsonResponse:
             else True
         )
 
-        slices, compilation_stages, compilation_text = compile_str(
+        slices, compilation_stages, compilation_text = Compilation().compile_str(
             request_data["circuit"], apply_litinski_transform
         )
         respnse_body = {
--- a/src/lsqecc/pipeline/lattice_surgery_compilation_pipeline.py
+++ b/src/lsqecc/pipeline/lattice_surgery_compilation_pipeline.py
@@ -14,7 +14,7 @@
 # License along with this library; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301
 # USA
-
+import json
 from typing import List, Optional, Tuple
 
 import qiskit.visualization as qkvis
@@ -42,33 +42,38 @@ def compile_file(
 def compile_str(
     qasm_circuit: str, apply_litinski_transform: bool = True
 ) -> Tuple[List[GUISlice], str]:
-    """Returns gui slices and the text of the circuit as processed in various stages"""
+    """Returns gui slices and string JSON of compilation text)"""
     composer_class = lscc.LatticeSurgeryComputation
     layout_types = lscc.LayoutType
 
     input_circuit = segmented_qasm_parser.parse_str(qasm_circuit)
 
-    compilation_text = "Input Circuit:\n"
-
-    compilation_text += qkvis.circuit_drawer(
+    input_circuit = qkvis.circuit_drawer(
         qkcirc.QuantumCircuit.from_qasm_str(qasm_circuit)
     ).single_string()
 
-    compilation_text += "\nCircuit as Pauli rotations:\n"
-    compilation_text += input_circuit.render_ascii()
+    circuit_as_pauli_rotations = input_circuit.render_ascii()
 
     # TODO add user flag
     input_circuit = input_circuit.get_y_free_equivalent()
 
+    circuit_after_litinski_transform = ""
     if apply_litinski_transform:
         input_circuit.apply_transformation()
         input_circuit = input_circuit.get_y_free_equivalent()
-        compilation_text += "\nCircuit after the Litinski Transform:\n"
-        compilation_text += input_circuit.render_ascii()
+        circuit_after_litinski_transform = input_circuit.render_ascii()
 
     logical_computation = llops.LogicalLatticeComputation(input_circuit)
     lsc = composer_class.make_computation_with_simulation(
         logical_computation, layout_types.SimplePreDistilledStates
     )
 
-    return list(map(sparse_lattice_to_array, lsc.composer.getSlices())), compilation_text
+    compilation_text = {
+        "input_circuit": input_circuit,
+        "circuit_as_pauli_rotations": circuit_as_pauli_rotations,
+        "circuit_after_litinski_transform": circuit_after_litinski_transform,
+    }
+
+    return list(map(sparse_lattice_to_array, lsc.composer.getSlices())), json.dumps(
+        compilation_text
+    )
--- a/src/lsqecc/pipeline/lattice_surgery_compilation_pipeline.py
+++ b/src/lsqecc/pipeline/lattice_surgery_compilation_pipeline.py
@@ -30,14 +30,16 @@ from lsqecc.resource_estimation.resource_estimator import estimate_resources
 GUISlice = List[List[Optional[vac.VisualArrayCell]]]  # 2D array of cells
 
 
+class Stage:
+    def __init__(self, name: str, body: str):
+        self.name = name
+        self.body = body
+
+
 class Compilation:
-    def stages(self) -> List[str]:
-        return [
-            "circuit",
-            "circuit_as_pauli_rotations",
-            "circuit_after_litinski_transform",
-            "resource_estimation",
-        ]
+    def stages(self, stages: List[Stage]) -> dict:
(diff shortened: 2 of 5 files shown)


Keelando:
This can stay open as a WIP, as merging it will momentarily break part of the UI, though it will be easy and quick to fix.

Can write formal tests for it though may require some guidance.

src/lsqecc/pipeline/lattice_surgery_compilation_pipeline.py
```diff
@@ -42,33 +42,58 @@ def compile_file(
...
+    """Returns gui slices and string JSON of compilation text)"""
     composer_class = lscc.LatticeSurgeryComputation
     layout_types = lscc.LayoutType
 
-    input_circuit = segmented_qasm_parser.parse_str(qasm_circuit)
-
-    compilation_text = "Input Circuit:\n"
-
-    compilation_text += qkvis.circuit_drawer(
+    input_circuit = qkvis.circuit_drawer(
```
gwwatkin: maybe worth calling `ascii_circuit`

src/lsqecc/pipeline/lattice_surgery_compilation_pipeline.py
```diff
@@ -42,33 +42,58 @@ def compile_file(
...
         qkcirc.QuantumCircuit.from_qasm_str(qasm_circuit)
     ).single_string()
 
-    compilation_text += "\nCircuit as Pauli rotations:\n"
-    compilation_text += input_circuit.render_ascii()
+    parsed_circuit = segmented_qasm_parser.parse_str(qasm_circuit)
+    circuit_as_pauli_rotations = parsed_circuit.render_ascii()
```
gwwatkin: Same here something with ascii

src/lsqecc/pipeline/lattice_surgery_compilation_pipeline.py
```diff
@@ -42,33 +42,58 @@ def compile_file(
...
 
     # TODO add user flag
-    input_circuit = input_circuit.get_y_free_equivalent()
+    parsed_circuit = parsed_circuit.get_y_free_equivalent()
 
+    circuit_after_litinski_transform = ""
```
gwwatkin: You can use:
```
circuit_after_litinski_transform : Optional[str]  = None
```

src/lsqecc/pipeline/lattice_surgery_compilation_pipeline.py
```diff
@@ -42,33 +42,58 @@ def compile_file(
...
     if apply_litinski_transform:
-        input_circuit.apply_transformation()
-        input_circuit = input_circuit.get_y_free_equivalent()
-        compilation_text += "\nCircuit after the Litinski Transform:\n"
-        compilation_text += input_circuit.render_ascii()
+        parsed_circuit.apply_transformation()
+        parsed_circuit = parsed_circuit.get_y_free_equivalent()
```
gwwatkin, line 62: Sorry for commenting this on your PR, but I was just going over this code with Alexandru today, a good name here would be to call this pauli_circuit or something like that

gwwatkin (commented):
Changes are good. If you are concerned with breaking the API, which is probably good because deploying these changes might be non trivial and might cause some downtime, here is what I suggest:

1. keep your changes
2. change `compilation_text` to `compilation_stages`
3. go to json pipeline and add compilation text back in there by `"\n".join`ing  `compilation_stages`'s 3 entries. The final JSON will have both `compilation_text` and `compilation_stages`
4. Merge this
5. Go to `web-ui` and add support for `compilation_stages`

gwwatkin:
@alexnguyenn  is `codecov/patch` failing because of `test_proportional_choice`?

gwwatkin:
I don't have permission to check

src/lsqecc/pipeline/lattice_surgery_compilation_pipeline.py
Keelando: I am not familiar at all with qiskit, so I am open to all suggestions.

My angle of attack on this was to fix it in an agnostic way, but I think it's good to have verbose and explicit variable names at every step with comments just for later reads of the code.

Keelando:
I'm not sure what the standards of tests we are using, but my idea was to just import the necessary modules, run the compile_str method which a particular circuit input, and just make sure it matches what we expect. Which is why you see commented circuit at the bottom, and the function call. I was having problems with relative python imports of modules in other places in the repo. Probably just a silly mistake somewhere. But python relative imports are more annoying than JS relative imports which are pretty intuitive. I would then just move the test to the appropriate directory

gwwatkin:
You can place your tests anywhere, as long as you run them with the python interpreter in the venv, you should be able to import the package as absolute, with `lsqecc`

src/lsqecc/pipeline/lattice_surgery_compilation_pipeline.py
Keelando: Do you know how the json encoder handles None in python?
Edit: looks like it will convert a None to a null... probably not a big deal.

gwwatkin (commented):
I think this is what we want, will look into getting codecov to pass without skewing the coverage results too much

src/lsqecc/pipeline/json_pipeline.py
```diff
@@ -77,8 +77,14 @@ def handle(json_request: str) -> JsonResponse:
             else True
         )
 
-        slices, compilation_text = compile_str(request_data["circuit"], apply_litinski_transform)
-        respnse_body = {"slices": slices, "compilation_text": compilation_text}
+        slices, compilation_stages, compilation_text = compile_str(
```
gwwatkin: Can also consider adding a `CompilatioResult` dataclass

Keelando:
@alexnguyenn Thoughts on disabling codecov on json_pipeline.py and lattice_surgery_compilation pipeline.py? These files basically call upon all the code in the repo, and even writing mock tests is a little beyond my expertise at the moment. I think we could circle back to adequate testing on these files, but it seems like possibly not an effective task at this exact moment, given that testing these files involves testing of much of the project. I have included a very basic, somewhat superficial, test on the names of the compilation stages so that if someone restructures the compilation output, it will fail until the stage names are added to the test.

alexnguyenn:
@Keelando you can merge even with codecov failing so I don't think it is a problem at all.

Keelando:
Thanks. It seems that I like to learn the hard way ;) but I did learn some things about python testing today.

gwwatkin (changes requested):
If @alexnguyenn  is fine with merging with codecov complaining so am I. Thanks @Keelando for looking into it.

Also like the the creation of the compilation class. I think it's a great idea that will allow us to add state to the computation which I see as desirable in the near future

I must however request some changes as the requirements are slightly different. Sorry @Keelando, this is a failure on my part for not explaining properly the rapidly changing situation.

With the new changes coming up in #235 there might be several intermediate steps, variably many and possibly an undetermined number ([like this](https://github.com/latticesurgery-com/lattice-surgery-compiler/blob/0b579aa740c3ab7621669d3291ee576e5264ddf9/debug/debug_contract_sk_result.py)). Also we might have alternate stages (like pyzx vs qiskit gate decomposition). So I think we are forced to create a list of stages.

If you create a data class like this:
```python
class Stage:
  def __init__(self, name:str, body: str):
    self.name = name
    self.body = body
```
and populate the list as you are now populating the dict. The JSON pipeline conversion should automatically handle converting to the right format and just by adjusting the type in web-ui (with no custom parsing!) you should have data in the desired format.

gwwatkin:
> If you create a data class like this:

This it to achieve two things, making stages flexible to be added and removed on this end, and making parsing trivial on web-ui

Keelando:
@gwwatkin Do you want a list of Stage objects that will be sent to the compilation class that turn on/off different parts of the compilation? Is that the idea? some of the compilation steps piggy back off each other. Is this Stage class just to populate the list of currently accepted compilation stages?
Edit: ended up deciding on a dictionary. At this point I am ready to hand it off, because I think you have a better idea of what changes are coming to the compiler. All I need for the UI is a clean JSON output of the different stages that doesn't need parsing.

src/lsqecc/pipeline/lattice_surgery_compilation_pipeline.py
```diff
@@ -30,14 +30,16 @@
...
 class Compilation:
-    def stages(self) -> List[str]:
-        return [
-            "circuit",
-            "circuit_as_pauli_rotations",
-            "circuit_after_litinski_transform",
-            "resource_estimation",
-        ]
+    def stages(self, stages: List[Stage]) -> dict:
+        stages_dict = {stage.name: stage.body for stage in stages}
```
gwwatkin, line 41: cool didn't know this syntax

gwwatkin (commented):
Ok @Keelando thanks for kick starting me I'll take over and will make a PR to the webui types. Will only be able to merge once the compiler is re deployed to lambda

Keelando:
Are there any new developments in terms of compilation stages you are adding? What is coded here does not anticipate everything that is to come but is an improvement I think.

gwwatkin:
Yes dropping the Pauli rotations all together in #257

gwwatkin:
Among other things

PR #221, opened 2022-01-25
Update README.md with new logo and badges

alexnguyenn:
As title. Feel free to remove/add badges as you see fit.


gwwatkin (approved):
Think it's good. Will probably want to add the universities at some point... but it will require cropping/scaling of the images. For now it's ok they are on the website anyways.

gwwatkin:
unless...

alexnguyenn:
> unless

wdym 😅

PR #222, opened 2022-01-25
Add backer's logos to readme

gwwatkin:
I was committing this right as you merged #221


alexnguyenn:
Right now it kinda look out of place for me (maybe cuz im using dark theme). Can you try a smaller size?

alexnguyenn (approved):

PR #224, opened 2022-01-26
fix to include isNegative attribute of measurements to llops

### LINKED ISSUE #164: [BUG] Take in account isNegative during simulation of measurements
Currently the isNegative flag is ignored but it has the effect of inverting the measurement outcome
https://github.com/latticesurgery-com/lattice-surgery-compiler/blob/c78c4fdfff645a8ca52c8a8fc2344f18d383aba0/src/lsqecc/simulation/logical_patch_state_simulation.py#L233-L245

isolatedinformation:
This includes changes to llops. Does not incorporate the changes in `simulation/`(currently interfering to #219!)
Partially fixes #164


gwwatkin (approved):
As @isolatedinformation mentioned it does not conflict with #219 since simulation does not change

Issue #227, opened 2022-01-28
[Feature Request] Basic resource estimation in the form of space time volumes


gwwatkin:
Issue Description
-----------------
Ultimately we want to know how many physical qubits and how long will it take to run a circuit.

Proposed Solution
-----------------

Start by outputting how many cells the program takes and how many slices. Enable defining a code distance

PR #229, opened 2022-01-28
Add basic resource estimation

gwwatkin:
Part of #227. Needs https://github.com/latticesurgery-com/web-ui/pull/117. 

Introduces some basic functionality to estimate how long and how many physical qubits are required to run the computation in question. Adds this data to the compilation text.

Including [OpenSurgery](https://github.com/alexandrupaler/opensurgery) as a git submodule to compute distillation box sizes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,9 +25,6 @@ jobs:
     - uses: actions/checkout@v2
       with:
         submodules: recursive
-    - name: Make OpenSurgery into a package
-      run: |
-        touch src/lsqecc/external/opensurgery/__init__.py
 
     - name: Set up Python 3.9
       uses: actions/setup-python@v2
@@ -51,9 +48,6 @@ jobs:
     - uses: actions/checkout@v2
       with:
         submodules: recursive
-    - name: Make OpenSurgery into a package
-      run: |
-        touch src/lsqecc/external/opensurgery/__init__.py
     
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,8 +26,8 @@ jobs:
       with:
         submodules: recursive
     - name: Make OpenSurgery into a package
-        run: |
-          touch src/lsqecc/external/opensurgery/__init__.py
+      run: |
+        touch src/lsqecc/external/opensurgery/__init__.py
 
     - name: Set up Python 3.9
       uses: actions/setup-python@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,7 +72,6 @@ jobs:
     
     - name: Test with pytest
       run: |
-        ls src/lsqecc/external/opensurgery
         if [[ ${{ matrix.python-version }} == '3.9' ]]; then
           pytest --cov=lsqecc --cov-report=term --cov-report=xml;
         else 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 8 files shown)


src/lsqecc/pipeline/lattice_surgery_compilation_pipeline.py
```diff
@@ -71,4 +72,7 @@ def compile_str(
         logical_computation, layout_types.SimplePreDistilledStates
     )
 
+    # TODO| when compilation stages are supported, remove the 'Circuit|' from the text
+    compilation_text += "\nCircuit| " + estimate_resources(lsc.composer.getSlices()).render_ascii()
```
gwwatkin: @Keelando , this line will conflict with #220, but it should be easy to fix. Ofc you can remove the comment and the `Circuit|` when merging 220. (probably want to keep the `\n` though)

gwwatkin: I'm still debating whether we should keep the compilation text, drop it or turn it into a "Log" stage. Because it's very useful to debug in plain text.

.github/workflows/test.yml
```diff
@@ -31,6 +31,7 @@ jobs:
         
     - name: Build package
       run: |
+        git submodule update --init --recursive
```
gwwatkin: @alexnguyenn  Looks like the CI is having problems fetching the submodules, thought this line would work ... do you have any ideas?

.github/workflows/test.yml
```diff
@@ -24,9 +24,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Sync submodules
+    - name: Update submodules
       run: |
-        git submodule sync --recursive
+        git submodule init
+        git submodule update --recursive
```
gwwatkin: Ok not sure what is going on here, this definitely work on my local machine

gwwatkin:
They really could try harder with error messages

gwwatkin:
Ok @alexnguyenn maybe you know what is going on ...  this makes pytest run on my machine on my machine, even with a newly cloned repo:
```
git submodule init
git submodule update --recursive
touch src/lsqecc/external/opensurgery/__init__.py
```
Though noticed that the checkout action does a recursive clone so the submodule part should not be necessary.

gwwatkin:
Ok @alexnguyenn ... made it eventually, please have a look at the github actions

.github/workflows/test.yml
```diff
@@ -23,7 +23,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    
+      with:
+        submodules: recursive
+    - name: Make OpenSurgery into a package
+        run: |
+          touch src/lsqecc/external/opensurgery/__init__.py
```
alexnguyenn:
```suggestion
    - name: Make OpenSurgery into a package
      run: |
        touch src/lsqecc/external/opensurgery/__init__.py
```
Alignment issue caused the failure I think

alexnguyenn:
I am confused here - what is this PR is doing and why are we adding opensurgery? Can you update the description with those infos?

I also just looked in the Github Actions log and seems like the test action didn't run at all.

gwwatkin:
@alexnguyenn, updated title and description. Had some problems with the github action ... squashed 10+ in which I was trying to fix it. The last one made no changes to the code, I think that's why the checks showed as not run, but you can see in the action log going a little back that they did go at one point before that (and many times they failed due to packaging problems :sweat_smile: ).

.github/workflows/test.yml
```diff
@@ -62,6 +72,7 @@ jobs:
     
     - name: Test with pytest
       run: |
+        ls src/lsqecc/external/opensurgery
```
alexnguyenn:
```suggestion
```
Don't think we need this.

alexnguyenn (approved):

PR #231, opened 2022-01-30
Remove Y's from all operators, not just pi/8 and measurements

gwwatkin:
As discussed in [this conversation](https://github.com/latticesurgery-com/lattice-surgery-compiler/pull/185/files/a80dc0ff52646d8595ada1f0c409b64698646de1#r767273177) there is nothing special about the pi/8 angle and we can remove Y operators from any angle.

The fact we were not running y removal on pi/4 was causing problems in circuits such as the one in the debug file, because Y operators were getting to the patch level.

### DIFF
--- a/src/lsqecc/pauli_rotations/circuit.py
+++ b/src/lsqecc/pauli_rotations/circuit.py
@@ -131,12 +131,7 @@ class PauliOpCircuit(object):
         y_free_circuit = PauliOpCircuit(self.qubit_num, self.name)
 
         for block in self.ops:
-            if isinstance(block, Measurement) or (
-                cast(PauliRotation, block).rotation_amount in {Fraction(1, 8), Fraction(-1, 8)}
-            ):
-                y_free_circuit.ops.extend(block.get_y_free_equivalent())
-            else:
-                y_free_circuit.add_pauli_block(copy.deepcopy(block))
+            y_free_circuit.ops.extend(block.get_y_free_equivalent())
 
         return y_free_circuit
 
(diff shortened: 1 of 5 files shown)


isolatedinformation (approved):

src/lsqecc/pauli_rotations/circuit.py
```diff
@@ -131,12 +131,7 @@ def get_y_free_equivalent(self) -> "PauliOpCircuit":
         y_free_circuit = PauliOpCircuit(self.qubit_num, self.name)
 
         for block in self.ops:
-            if isinstance(block, Measurement) or (
-                cast(PauliRotation, block).rotation_amount in {Fraction(1, 8), Fraction(-1, 8)}
-            ):
-                y_free_circuit.ops.extend(block.get_y_free_equivalent())
-            else:
-                y_free_circuit.add_pauli_block(copy.deepcopy(block))
+            y_free_circuit.ops.extend(block.get_y_free_equivalent())
```
isolatedinformation, line 134: would it make sense to implement `if` block.has_y()` before calling `y_free_equivalent()`? Or move have this if check at the top of the function?

isolatedinformation (approved):

src/lsqecc/pauli_rotations/circuit.py
gwwatkin: Yeah good point. there is a check inside PauliProductOperator that scans for ys anyways.

From my preliminary data on profiling, it turns out that a major bottleneck in the compilation are deep copies so we'll have to work on fixing those. But I think we can leave this as an optimization for then, if it comes up the profiling that these y_free-related deep copies are taking time. It seems to be a problem mostly with patch computation

gwwatkin:
CC @alexnguyenn

Issue #233, opened 2022-02-01
[BUG] Cannot decompose arbitrary Z rotations to Clifford+T


gwwatkin:
Issue Description
-----------------
For example, cannot decompose a pi/16 rotation about Z. This makes it impossible to run a QFT, making it difficult to run any useful benchmark (like in #228)

How to Reproduce
----------------

1. Cherry-pick [this commit](https://github.com/latticesurgery-com/lattice-surgery-compiler/commit/45f84cbd0670c2eafeff3e972f8af3f554dcf12f) to fix the fact that pyzx has trouble parsing circuits of a certain form
2. run a circuit with a crz gate (qft example from [benchmark repo](https://github.com/latticesurgery-com/benchmark) below, but any rz gate will do)

### Code Snippet

```python

        slices, text = lscp.compile_str(decomposed_qft.qasm(),
                                        simulation_type=lscp.lssim.SimulatorType.NOOP,
                                        apply_litinski_transform=False)
```

The circuit is:
```qasm
Decomposed raw QASM input:
OPENQASM 2.0;
include "qelib1.inc";
qreg q0[3];
u2(0,pi) q0[2];
barrier q0[0],q0[1],q0[2];
rz(pi/4) q0[2];
cx q0[1],q0[2];
rz(-pi/4) q0[2];
cx q0[1],q0[2];
u2(0,pi) q0[1];
barrier q0[0],q0[1],q0[2];
rz(pi/4) q0[1];
rz(pi/8) q0[2];
cx q0[0],q0[2];
rz(-pi/8) q0[2];
cx q0[0],q0[2];
cx q0[0],q0[1];
rz(-pi/4) q0[1];
cx q0[0],q0[1];
u2(0,pi) q0[0];
barrier q0[0],q0[1],q0[2];
```

### Error Output

```
  File "/home/george/projects/latticesurgery-com/benchmark/qft.py", line 35, in <module>
    slices, text = lscp.compile_str(decompose_qft.qasm(),
  File "/home/george/projects/latticesurgery-com/benchmark/external/lattice-surgery-compiler/src/lsqecc/pipeline/lattice_surgery_compilation_pipeline.py", line 77, in compile_str
    logical_computation = llops.LogicalLatticeComputation(input_circuit)
  File "/home/george/projects/latticesurgery-com/benchmark/external/lattice-surgery-compiler/src/lsqecc/logical_lattice_ops/logical_lattice_ops.py", line 100, in __init__
[...]

PR #234, opened 2022-02-01
Fix parametrized gates breaking PyZX parsing

gwwatkin:
Gates such as:
```qasm
rz(-pi/8) q[0];
```
once parsed into AST and then rendered to be passed to PyZX, become
```qasm
rz(-(pi/8)) q[0];
```
While this is valid QASM, PyZX's parser hard codes the parsing of angles. To avoid crashing PyZX's parser we re-parse the circuit in qiskit so that angle expressions are simplified.

### DIFF
--- a/src/lsqecc/pauli_rotations/segmented_qasm_parser.py
+++ b/src/lsqecc/pauli_rotations/segmented_qasm_parser.py
@@ -85,11 +85,7 @@ class _SegmentedQASMParser:
         gates_as_qasm = qiskit.qasm.node.Program(segment).qasm()
 
         text_qasm_program_wrapper = (
-            "OPENQASM 2.0;\n"
-            + 'include "qelib1.inc";\n'
-            + self.qreg.qasm()
-            + "\n"
-            + gates_as_qasm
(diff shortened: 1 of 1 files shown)


gwwatkin:
pre-commit.ci autofix

src/lsqecc/pauli_rotations/segmented_qasm_parser.py
```diff
@@ -82,16 +82,22 @@ def segment_to_circuit(self, segment: Segment):
...
             + 'include "qelib1.inc";\n'
             + self.qreg.qasm()
             + "\n"
-            + program_wrapper_node.qasm()
+            + gates_as_qasm
         )
+        # Feeding text_qasm_program_wrapper directly to pyzx qould cause it to complain because
+        # gates_as_qasm, being generated directly from the AST, has extra parentheses pyzx doesn't
+        # like. So we feeding it back to qiskit we remove those extra parentheses
+        # TODO try avoid this step by going directly to circuit
```
gwwatkin, line 93: Probably not worth addressing this TODO because, in light of #233 we might have to drop PyZX

alexnguyenn (approved):

PR #235, opened 2022-02-02
Break down rotations other than pi/2, pi/4, pi/8

gwwatkin:
Break down rotations about X or Z up to pi/2^31.

Along the way separated the code to convert from gate to Pauli Rotations ad added some tests.

Part of #233. Doesn't close it because it doesn't handle arbitrary rotations, but pi/2^n should get us through many benchmarks, which are higher priority. Handling more arbitrary gates is just a matter of either caching more rotations, embedding gidsynth or decomposing a pi-angle as a sum of negative powers of 2.


src/lsqecc/pauli_rotations/rotation.py
```diff
@@ -218,6 +218,32 @@ def __eq__(self, other) -> bool:
...
+        """Get an approximation in terms of pi/2, pi/4 and pi/8"""
+        # TODO use SK
+        axis = list(filter(lambda op: op != PauliOperator.I, self.ops_list))
+        if len(axis) != 1:
+            raise Exception("Can only approximate single qubit rotations")
+
+        if axis == PauliOperator.X:
+            raise NotImplementedError(f"decompose X rotation by {self.rotation_amount}")
+        elif axis == PauliOperator.Z:
+            raise NotImplementedError(f"decompose Z rotation by {self.rotation_amount}")
```
gwwatkin: @isolatedinformation this is where the code to break down arbitrary rotations will go. We could have pre built tables for Common pi/2^k Z rotations

gwwatkin:
CC @alexnguyenn for the refactoring

gwwatkin:
@isolatedinformation  and @alexnguyenn I suggest reviewing the changes as they are now ... There will pobably be many other before we can merge this. To avoid having a massive diff it's probably best to start reviewing

gwwatkin:
Can be reviewed commit by commit

src/lsqecc/pauli_rotations/rotation.py
```diff
@@ -280,6 +280,57 @@ def from_list(pauli_ops: List[PauliOperator], rotation: Fraction) -> "PauliRotat
...
 
+    @staticmethod
+    def from_fractional_phase(
+        num_qubits: int, target_qubit: int, phase_type: PauliOperator, phase: Fraction
+    ):
+        return PauliRotation.from_list(
+            [PauliOperator.I] * target_qubit
+            + [phase_type]
+            + [PauliOperator.I] * (num_qubits - target_qubit - 1),
+            rotation=phase / 2,
```
isolatedinformation, line 316: I thought we were using `P_phi = exp(-i*P*phi)` as the convention from Litinski's convention, not the `P_phi = exp(-i*P*phi*0.5)`

src/lsqecc/pauli_rotations/rotation.py
```diff
@@ -280,6 +280,57 @@ def from_list(pauli_ops: List[PauliOperator], rotation: Fraction) -> "PauliRotat
             r.change_single_op(i, op)
         return r
 
+    @staticmethod
+    def from_fractional_phase(
```
isolatedinformation: might help a change of name, since this makes a `PauliRotation` as wide as the circuit, or just docs to indicate that

src/lsqecc/pauli_rotations/rotation.py
gwwatkin: Phase is the argument of the `rz` gate, my understanding was that the argument is the relative phase to the `|1>` basis state, so `rz theta` = `exp(-i*pi*theta/2)` ... but this is definitely a subtle issue and a nice catch.

So what we are doing now is this:
```
Input Circuit:
     ┌─────────┐
q_0: ┤ Rz(π/8) ├
     └─────────┘
Circuit as Pauli rotations:
 q0---|Z|
pi*  1/16
```

Do you think it's correct?

isolatedinformation: In qasm, the `theta/2` definition is used and internal in the `PauliRotation` class, we use the `theta`. This circuit is performing the correct operation, but just that generally `theta/2` definition is used. Line 291 takes care of the convention to maintain correctness of calculation. However, the end user might be confused, because he input `pi/8` but saw `pi/16` when the circuit is presented as Pauli Rotations

gwwatkin: Ok ... well it seems that the standard is to keep `theta` in the Pauli formalism ... but I do see how this would be confusing to users

.pre-commit-config.yaml
```diff
@@ -1,7 +1,7 @@
 # See https://pre-commit.com for more information
 
 default_language_version:
-  python: python3.9
+  python: python3.10
```
gwwatkin:
```suggestion
  python: python3.9
```

src/lsqecc/pauli_rotations/rotation.py
gwwatkin: Added coments in various places to try make this issue more evident and what to do clearer, but I don't think we want to change our Pauli formalism (which matches Litinski's and [others's](https://arxiv.org/pdf/2109.02746.pdf)) based on what qiskit's and gridsynth's rz convention is

src/lsqecc/pauli_rotations/rotation.py
gwwatkin: renamed to make it clear that it is starting from a single qubit gate

gwwatkin:
@isolatedinformation  and @alexnguyenn  ready to merge.

Not sure what is up with the doc build, I don't think it depends on my changes.

gwwatkin:
I see some warnings but no errors in the doc build

alexnguyenn:
yeah I am confused as well, not sure what is going on with docs building.

src/lsqecc/pauli_rotations/rotation.py
```diff
@@ -270,9 +270,10 @@ def get_basic_form_decomposition(self) -> List["PauliRotation"]:
             fractions_with_unit_numerator = decompose_pi_fraction(self.rotation_amount)
             output_rotations = []
             for unit_rotation_amount in fractions_with_unit_numerator:
-                new_rotation = copy.deepcopy(self)
-                new_rotation.rotation_amount = unit_rotation_amount
-                output_rotations.append(new_rotation)
+                if unit_rotation_amount.denominator != 1:
```
isolatedinformation, line 287: Why is this condition enforced? I am not able to follow

isolatedinformation (approved):

src/lsqecc/pauli_rotations/rotation.py
gwwatkin: We discard rotations by `pi` because they have no effect

gwwatkin:
```suggestion
                # Discard rotations by pi because they have no effect
                if unit_rotation_amount.denominator != 1:
```

gwwatkin:
@alexnguyenn looks like the problem is now that the doc builds in a different directory so it can't fin files in assets. 
```
FileNotFoundError: [Errno 2] No such file or directory: './assets/cached_phase_rotation_approximations.txt'
```
Maybe the issue is that we need to copy over assets? like in `setup.(py|cfg)`.

I would really like to have this merged and deployed soon because it makes a big difference ... now we can compile QFT circuits

gwwatkin:
@alexnguyenn got the doc build to pass, so merging, a lot of changes piling up onto this PR, want to keep things moving.

PR #239, opened 2022-02-04
Add python 3.10 to tests

gwwatkin:
Works on my machine. @alexnguyenn should we drop 3.10? I was kind of glad we had it when we experimented wit pypy ... but that was a failure eventually.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,7 +42,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
 
     steps:
     - uses: actions/checkout@v2


alexnguyenn:
Also I think we should start moving to [PEP 585](https://www.python.org/dev/peps/pep-0585/) (not using `from typing import ...`). For 3.10 I am interested in pattern matching and maybe using | for union type - is there any feature from it you want to adapt?

alexnguyenn:
> should we drop 3.10

Wait do you mean the reverse?

gwwatkin:
> For 3.10 I am interested in pattern matching and maybe using | for union type - is there any feature from it you want to adapt?

Yes I'm interested in pattern matching as well (*Haskell vibes* :scream:).  Though not from 3.10, I wish we had TypeAlias. I think some of the older versions don't support it.

> Wait do you mean the reverse?

Yeah, sorry drop 3.7.

Though before we move completely to 3.10 lets check that it runs fine on the lambda (I don't see why it wouldn't though)

alexnguyenn (approved):

alexnguyenn:
> Yeah, sorry drop 3.7.

I think it is ok to leave it there right now as the tests are still passing.

alexnguyenn:
https://github.com/latticesurgery-com/lattice-surgery-compiler/blob/ad3c77ad1ea93969dfc12e445c6a07fb723a4547/.github/workflows/test.yml#L75-L80

Also do you want the coverage report to be generated for 3.9 or 3.10

gwwatkin:
> Also do you want the coverage report to be generated for 3.9 or 3.10

Making a second pr to change a configs to 3.10

PR #244, opened 2022-02-06
 Fast Slice Composer in C++

gwwatkin:
Setup the `cmake`, `pybind` stack to build. Still have to make it usable from python's `lsqecc` but you can do a trivial test by cd`ing to the build directory an manually importing the complied shared object.

No functionality yet, but started thinking about data structures.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ __pycache__/
 
 # C extensions
 *.so
+*.a
 
 # Distribution / packaging
 .Python
@@ -15,7 +16,6 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
 lib64/
 parts/
 sdist/
@@ -94,3 +94,8 @@ cython_debug/
 .idea
 *.iws
 out/
+
+### CMAKE Files ###
+CMakeFiles/
+*.cmake
+CMakeCache.txt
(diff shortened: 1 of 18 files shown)


alexnguyenn:
Just wondering can't you just `pip install pybind`?

gwwatkin:
> Just wondering can't you just `pip install pybind`?

I tried, it is indeed possible to do it, but then you have to go chase down the header files from the venv and either add that to your include path or copy them to your project to make them work with the IDE. The pybind example using pip/setuptools is incompatible with CLion so we'd need a dual build system with cmake anyways.

gwwatkin:
If I see we are making good progress with this experimental branch, I'll use a cleaner solution. I could set CMake to point to the the venv's site-packages with the headers so we can eliminate the need to have the submodule (note that that's the way they sugest to use [pybind11 with cmake](https://github.com/pybind/cmake_example)). And then schedule cmake to run after pybind is installed.

Another alternative might be Conan, that's how qiskit-aer manages c++ dependencies

gwwatkin:
Something like Conan would be useful because we want to be able to add other packages like Eigen

gwwatkin:
Moved to it's own repo [liblsqecc](https://github.com/latticesurgery-com/liblsqecc). We probably won't even need to interface python with the introduction of the LS Instruction format #246

gwwatkin:
Though we might still want python bindings to expose as an API :thinking:

PR #245, opened 2022-02-08
Stub of resource estimation from Logical Lattice Ops

gwwatkin:
Changes to make the resource estimator work from Logical Lattice Ops (similar to topological assembly). 

Pr on `clifford+t` branch because need changes from it to make it operable with pi/16 gates.


gwwatkin:
Moved to #248 as #235 was merged

PR #249, opened 2022-02-09
Typo in arbitrary approximation file's comment

gwwatkin:
It's already confusing enough without the typo...

### DIFF
--- a/src/lsqecc/pauli_rotations/pi_over_2_to_the_n_rz_gate_approximations.py
+++ b/src/lsqecc/pauli_rotations/pi_over_2_to_the_n_rz_gate_approximations.py
@@ -1,6 +1,6 @@
 # Gate sequences to approximate rz gates with arguments ^n.
 # Key is n so 0->rz(pi/1), 1->rz(pi/2), 2->rx(pi/4) ...
-# Note that rz(theta) is a rotation by rx(theta/2) in our convention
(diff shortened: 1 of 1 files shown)


src/lsqecc/pauli_rotations/pi_over_2_to_the_n_rz_gate_approximations.py
```diff
@@ -1,6 +1,6 @@
 # Gate sequences to approximate rz gates with arguments ^n.
 # Key is n so 0->rz(pi/1), 1->rz(pi/2), 2->rx(pi/4) ...
-# Note that rz(theta) is a rotation by rx(theta/2) in our convention
+# Note that rz(theta) is a rotation by theta/2 about the Z axis in our convention.
```
isolatedinformation, line 3: which convention is this ? The one by Litinksi or the generic convention ?

gwwatkin: It also is the same as Litinski, I don't know what is the generic convention? it seems like also in that AWS paper you sent me the use the same convention. `rz` is just qiskit's gate, and `R_z` which happens to be what gridsynth uses

isolatedinformation (approved):

PR #254, opened 2022-02-11
Stub of Manual QASM Parser

gwwatkin:
Parsing with both PyZX and QASM breaks for angles smaller than `pi/2^30..pi/2^48`. This was becoming a problem, in benchmarking with Larger QFTs. So here is a parser that can read such circuits and interpret them as Pauli Rotations.

Draft because it's still missing tests, but I'm using it already. Making it available so @alexnguyenn you can start reviewing. 

For the time being, will not be part of the user pipeline on latticesurgery.com. But I would like to merge it so it can be used in debug files and other projects.

@alexnguyenn, since you've mentioned it before, you might be interested in completing this parser and making it more robust so that it can fully replace pyzx/qiskit (we can handle most gates pyzx can handle now).


debug/debug_manual_qasm_parse.py
```diff
@@ -0,0 +1,18 @@
...
+
+"""
+
+
+
+if __name__ == "__main__":
+    qasm = qft_benchmark.qft_gate(4).qasm()
+    print()
+
+    print(PauliOpCircuit._manual_parse_from_reversible_qasm(qasm))
```
gwwatkin:
```suggestion

if __name__ == "__main__":
    qasm = qft_benchmark.qft_gate(4).qasm()
    print(PauliOpCircuit._manual_parse_from_reversible_qasm(qasm))
```

src/lsqecc/pauli_rotations/rotation.py
```diff
@@ -366,6 +368,29 @@ def from_cz_gate(
...
+        Z = PauliOperator.Z
+
+        return (
+            [
+                PauliRotation.from_r_gate(num_qubits, control_qubit, Z, phase / 2),
+                PauliRotation.from_r_gate(num_qubits, target_qubit, Z, phase / 2),
+            ]
+            + PauliRotation.from_cnot_gate(num_qubits, control_qubit, target_qubit)
+            + [PauliRotation.from_r_gate(num_qubits, target_qubit, Z, -phase / 2)]
+        )
```
gwwatkin, line 392: @isolatedinformation, the identity you showed me

isolatedinformation: https://www.nature.com/articles/s41598-018-23764-x.pdf?origin=ppub, just putting the paper here, cus I keep losing it

alexandrupaler (approved):
Ok with it, after trusting George

src/lsqecc/pauli_rotations/rotation.py
alexandrupaler: @isolatedinformation Of course, but George did not trust my circuit identity skills :)

PR #260, opened 2022-02-16
Make lsqecc pep 561 compliant

mstechly:
Imagine some other projects wants to use `lsqecc`. They can use it, but unfortunately the won't be able to run mypy against any imports from `lsqecc`, unless package is [PEP 561 compliant](https://www.python.org/dev/peps/pep-0561/).

This PR adds the components necessary to make it works.

It also solves a couple of other issues:
1. Due to the configuration in `setup.cfg`, when I was building `lsqecc` locally and trying to install with wheel, I noticed that the package is empty. Changing to `packages = find:` in `setup.cfg` solved this issue.
2. I've also added `zip_safe=False`, as it's known to interfere with editable installations (see [here](https://github.com/python/mypy/issues/7508#issuecomment-531965557)
3. `namespace_packages=true`  – this will allow to avoid `__init__.py` at all the levels. I'm not sure if this change is actually desirable, please let me know.
4. `explicit_package_bases=true` allows to avoid some issues with mypy configuration in the future.

If you want I can add inline comments to those config files explaining the reason behind them.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include src/lsqecc/py.typed
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,9 +12,6 @@ line-length = 100
 
 [tool.mypy]
 ignore_missing_imports = true
-namespace_packages = true
-explicit_package_bases = true
-mypy_path = "src"
 
 [tool.pytest.ini_options]
 testpaths = ["tests"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,6 +12,9 @@ line-length = 100
 
 [tool.mypy]
 ignore_missing_imports = true
+namespace_packages = true
+explicit_package_bases = true
+mypy_path = "src"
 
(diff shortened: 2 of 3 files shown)


gwwatkin (approved):
Thanks @mstechly. I think all these changes are desirable if they make it easier for other people to use the package. For the `__init__.py` thing, I think that since all our sources are in the `src` directory, everything that is not a package will be out of it, which sounds to me like we don't init files at all levels.

@alexnguyenn might have more insight than I do into packaging though.

alexnguyenn (approved):
Thank you for the PR. I agree that these changes are desirable. 
* For `namespace_packages=true`, I don't think we are using any namespace packages atm, but I am ok with having it there just in case (as long as there is no conflict). 
* It will be great if you could put some comments in as well.

mstechly:
@alexnguyenn I have added some comments and removed some of the options after consideration. It should work fine and I guess it's better to add them when we need them, rather than get into trouble because of having them and not understanding all the consequences.

Issue #246, opened 2022-02-08
[Feature Request] Come up with an instruction set for lattice surgery that is layout independant


gwwatkin:
Issue Description
-----------------
Talking with @alexandrupaler, it came up that it is very desirable to have a language to express lattice surgery instruction before layout and routing.

Proposed Solution
-----------------
The following instructions, new line separated. Note that each instruction (except for the qubit declaration) also have a numeric opcode to reduce file size)

```
DeclareLogicalQubitPatches <patch_id>,<patch_id>,... // must be the first instruction
(Init|0) <patch_id> ("|0>" | "|+>" | + | 0)
(RequestMagicState|1) <patch_id>
(MultiBodyMeasure|2) <patch_id>:(X|Z),<patch_id>:("X"|"Z"),...
(MeasureSinglePatch|3) <patch_id> (X|Z)
(LogicalPauli|4) <patch_id> (X|Z)
(SGate|5) <patch_id>
(HGate|6) <patch_id>
(RequestYState|7) <patch_id>
```

Where`<patch_id>` is a 32 bit unsigned integer.

### Reference implementation
* Writing in the [`__repr__` methods of LS-Instructions](https://github.com/latticesurgery-com/lattice-surgery-compiler/blob/dev/src/lsqecc/ls_instructions/ls_instructions.py) and the[ short-hand version ](https://github.com/latticesurgery-com/lattice-surgery-compiler/pull/266).
* Parsing in [liblsqecc#4](https://github.com/latticesurgery-com/liblsqecc/pull/4/files#diff-5671d05f60e80224280085a289783757458bebf0e33f453121c2e9882c5d19aa)

### Corrective terms/ Conditional Extension
A problem for which we are still discussing a solution for are corrective terms. LS-Instructions need corrective terms when the measurement reads the negative eigenvalue. E.g. from Litinski for pi/8, but the same applies to all teleported gates via multibody measurement:
![image](https://user-images.githubusercontent.com/36427091/153034112-976fe78b-908a-44c0-9b09-7be94c77d7d2.png)
 * Ideally, these corrective terms would just be tracked by updating the Pauli frame. In this case all we have to assume is that an implementation reading from this protocol will automatically know how to update the Pauli frame based on measurement outcomes. However, this is not trivial because, as shown above, the sequence of corrections is not immediate and is making assumptions about what the operation from which the multibody measurements come from is. (a pi/8 rotation in the example)
[...]

PR #266, opened 2022-03-06
LS Instruction short-hand codes

gwwatkin:
Adds a class to print [LS-Instructions](https://github.com/latticesurgery-com/lattice-surgery-compiler/issues/246) with short-hand codes instead of the full opcode, i.e:
```
RequestMagicState => 1
MultiBodyMeasure => 2
MeasureSinglePatch => 3
...
```

Takes the file size of LS-Instruction files down drastically since most of the file size was taken up by the long opcodes.

##### Benchmark
QFT 100:
```
Generate LS Instructions
Generated: 39101029 lines, 761581272 chars for full length mnemonics and 346705421 for shorthand LS
```
that is 761.6 MB vs 346.7 MB

### DIFF
--- a/src/lsqecc/ls_instructions/ls_instructions.py
+++ b/src/lsqecc/ls_instructions/ls_instructions.py
@@ -37,19 +37,12 @@ class Init(LSInstruction):
 
 
 @dataclass
-class RequestMagicState(LSInstruction):
+class MeasureSinglePatch(LSInstruction):
     patch_id: PatchId = -1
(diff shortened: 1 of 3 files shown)


alexnguyenn:
@gwwatkin can you put in more details on what you did that results in the decrease?

alexnguyenn (approved):

gwwatkin:
@alexnguyenn Specified the change and updated the issue for the spec #246

Issue #263, opened 2022-02-22
[Feature Request] Better QFT Compilation in the Pauli formalism 


gwwatkin:
Issue Description
-----------------

Our implementation of the Pauli Formalism is very slow - see the benchmark in #258. Additionally it also generates 4x more ls instructions, but there is a chance to generate even fewer.

Proposed Solution
-----------------

#### Compilation speed

The slowness, compared to the gate based pipeline, is due to the fact that QFT circuits are dominated by very long single qubit Clifford+T approximations and our pipeline pads single qubit rotations with identities to span the whole width of the circuit. Switching to a sparse format (like in the gate based pipeline) is likeley to speed up this step and drastically reduce memory footprint.

#### Computation depth
The bottle neck is approximations of arbitrary rotations, where we are going to gates and back, Increasing the length at the computation at every step:

Now, our pipeline is:
```
rotations -> approximation with gates -> rotations -> instructions
```
We have the opportunity of using directly Pauli rotations to approximate rotations: 
```
goes rotations -> approximation with rotations -> instructions.
```
The approximation with rotations translates directly to LS instructions. With this pipeline, we should generate fewer ls instructions than with the gate based pipeline, which still does the Hadamards and separates groups like `ZST` which are a single `Z_7pi/8` rotation

To do that, we reinterpret the Gridsynth approximations, which are in a normal form (as shown below), and cache them directly as rotations. We do so by grouping Zs, Ss and Ts and "sandwiching" them between Hadamard gates, and seeing that as a basis change:
Some gate -> `TSHTHTHSTHZSTHTH...`
 -> `TS` `HTH` `T` `HSTH` `ZST` `HTH` ...
 -> `Z_3pi/8`  `HZ_pi/8H` `Z_pi/8` `HZ_3pi/6H` `Z_5pi/8` `HZ_pi/8H`...
 -> `Z_3pi/8`  `X_pi/8` `Z_pi/8` `X_3pi/6` `Z_7pi/8` `X_pi/8` ...

Then these translate directly to ls instructions.

Additional References
---------------------

If applicable, provide some references that will help us better understand the request.

isolatedinformation:
What is the status of this issue?

gwwatkin:
@isolatedinformation No work has been done yet. As by our conversation on Slack, passing it over to you.

CC: @alexnguyenn

isolatedinformation:
[...]

PR #269, opened 2022-03-17
Better QFT Compilation python prototyping

isolatedinformation:
This PR solves #263 of Better QFT compilation in the Pauli Formalism
- Partitions a long gate sequence to strings that can be compressed as a single Pauli rotation (`gates/compress_rotation_approximations.py`)
For instance, processing the gate with phase = `pi/2^100`  gives the following, where each partition is one PauliRotation gate:
```bash
The number of partitions is 466
The number of gates is 1172
```
- incorporate the compression in `PauliRotation`
- [x] Testing
	- [x] 	`partition_gate_sequence`
	- [x] `to_basic_form_approximation_and_compression`


isolatedinformation:
@gwwatkin Do we want to cache the PauliFrame rotations also? A lot of the sequences are either `HSTH` or `HTH`!

isolatedinformation:
compression_X_rotations


compression_Z_rotations

gwwatkin (approved):
Can't go too thoroughly right now but ok with merging to go ahead since it doesn't affect the main pipeline.

We need to check thas LLOPs work as expected with unusual rotations like 5pi/8

Could you remove the images from the assets and then do a use the squash merge button? It's so we don use up the git history space with files. You can add the images to the PR on github.

isolatedinformation:
Done! I will raise an issue for the LLOPS checking

Issue #272, opened 2022-03-19
[Feature Request] Sparse Representation of operations in the Pauli Formalism


isolatedinformation:
Issue Description
-----------------
Our pipeline pads single-qubit rotations with identities to span the whole width of the circuit. Switching to a sparse format (like in the gate-based pipeline) is likely to speed up this step and drastically reduce memory footprint. (as pointed out in #263 )

<add more details as needed>

gwwatkin:
I suggest looking at how gates are implemented as an example. Could even add the pauli rotation there, similarly to how the rz is implemented (inheriting from Gate), but with one more parameter for the axis (X/Z)

alexnguyenn:
Should we save this for when we start migrating to C++ @gwwatkin?

gwwatkin:
@alexnguyenn and @isolatedinformation, would be nice to get this done in python so we can squeeze it into the paper. It should be fairly quick to implement

isolatedinformation:
https://github.com/latticesurgery-com/lattice-surgery-compiler/blob/d59f4472fe9abfdb303017e9938f9fec73f269d1/src/lsqecc/gates/gates.py#L53-L64

@gwwatkin shouldn't line 62 be the T gate and not the S gate?

gwwatkin:
Yes you are right! Good catch

isolatedinformation:
added the fix for this in #273

gwwatkin:
@isolatedinformation I think we can can close this after [this comment from 274](https://github.com/latticesurgery-com/lattice-surgery-compiler/pull/274#pullrequestreview-919939757) is adressed:
> One last think to consider is the half angle conversion for Pauli Rotations. Right now all the angles are double those in the convention of the latest lattice surgery papers and our existing block based Pauli rotations. I think we should also switch here to half angles

PR #274, opened 2022-03-23
Sparse representation of PauliRotations in the Gate formalism

isolatedinformation:
Sparse rotations have been implemented with support for compressions of Cliffords to Rotations! 
Hopefully solves #272 
Benchmarking with QFT is yet to be done


src/lsqecc/gates/gates.py
```diff
@@ -49,26 +55,59 @@ class T(Gate):
     pass
 
 
+@dataclass
+class PauliRotations(Gate):
+    phase: Fraction = Fraction(1, 1)
```
gwwatkin, line 60: A detail about consistency when working with Pauli Rotations. In the existing Pauli formalism we refer to the angle of rotation as the half angle a-la-Litinksi/AWS so `S=Z_{pi/4}`. It seems that here we are departing from that convention, we are using the phase in the exponent of the `|1><1|` term. 

I think we should consider carefully if we want to make this change.

src/lsqecc/gates/gates.py
```diff
@@ -49,26 +55,59 @@ class T(Gate):
...
+    def to_clifford_plus_t(self, compress_rotations: bool = False) -> Sequence["Gate"]:
+        if self.axis == PauliOperator.Z:
+            if self.phase == Fraction(1, 1):
+                return [Z(self.target_qubit)]
+            elif self.phase == Fraction(1, 2):
+                return [S(self.target_qubit)]
+            elif self.phase == Fraction(1, 4):
+                return [T(self.target_qubit)]
+            else:
+                return approximate.approximate_rz(RZ(phase=self.phase), compress_rotations)
```
gwwatkin, line 72: We've got to be careful that this doesn't get called more again on the approximated sequences, otherwise we loose the advantage of grouping gates.

tests/gates/snapshots/approximate_test/test_approximate_rz_with_compression/phase0/gates.txt
```diff
@@ -0,0 +1 @@
+[]
```
gwwatkin: I don't think this should be empty

gwwatkin: I would expect it to be `[PauliRotations(target_qubit=1, phase=Fraction(1, 1), axis=Z)]`

gwwatkin (changes requested):
It's great you got this up quickly. This is a move in the right direction but I think we need some refinements.

Some tests are appearing to be empty when they shouldn't be. I think the problem is with the logic in `partition_gate_sequence` which was pushed in the previous PR #269. 
I propose fixing it with something like this https://replit.com/@gwwatkin/partitiongatesequence

Also what should the behavior be for strings like `SSSX`? should it be up to `partition_gate_sequence` to split this? or the from gate string function? 

In general it would be nice to see some small non snapshot tests to test the over all correctness and define these behaviors.

Sorry, these were probably comments for the previous PR, but I wasn't able to review it properly.

src/lsqecc/gates/gates.py
isolatedinformation: I checked the logic for this and it was not being called, with the default value being `False` and we do not call the `to_clifford_plus_t` again! Did you find it otherwise?

src/lsqecc/gates/gates.py
isolatedinformation: I assumed in the Gate Formalism, we are following the exponent convention! I just took inspiration from the RZ gate definitions!

isolatedinformation:
And one more thing for which support needs to be added is parsing qasm into `PaulIRotations`! I have not touched the parsing code!

isolatedinformation:
For the behavior of `SSSX`, the best would be to break it down as `SSS` and `X`, which `partition_gate_sequence` does now!

isolatedinformation:
there are non-snapshots tests that are there for `partition_gate_sequence` in `gates/compress_rotation_approximations_test.py` :)  
With regards to whether the breakage of strings `SSSX` should go to `from_gate_string`, well it can! Right now, `partition_gate_sequence` breaks down everything that is not on the same basis and `from_gate_string` assumes the strings it receives are already passed through the `partition_gate_sequence` function!  I pushed some changes now with some bug fixes, but below are some examples that explain the expected behavior:
"SSSXHSTHTHSTHTH" --> ["SSS", "X", "HSTH", "T", "HSTH", "T", "H"]
("STXH", ["ST", "X", "H"]), 
("HSTX", ["H", "ST", "X"]),
("XHST", ["X", "H", "ST"]),

If the behavior is too ambiguous, let's make any design changes necessary that are necessary

tests/gates/snapshots/approximate_test/test_approximate_rz_with_compression/phase0/gates.txt
isolatedinformation: It does now! The bug was in `partition_gate_sequence` giving an empty string for the indices 0,1 and 2

src/lsqecc/gates/gates.py
gwwatkin: :+1:

gwwatkin: No I haven't check, guess we would find out if rotations expand again. In general though I might replace this return with an `assert False`, just so that we avoid chasing down complicated logic.

gwwatkin:
> And one more thing for which support needs to be added is parsing qasm into `PaulIRotations`! I have not touched the parsing code!

For the purposes of the QFT benchmark we don't need to since all gates are CNOTs, Hadamards (which we don't want to break down anymore because we have them transversally) and rotations

src/lsqecc/gates/compress_rotation_approximations.py
```diff
@@ -8,6 +8,13 @@ def partition_gate_sequence(gate_approximation: str):
     for index, gate in enumerate(gate_approximation):
         if gate in {"S", "T"}:
             partition += gate
+            if index == len(gate_approximation) - 1:
+                if not in_x_basis:
+                    partitioned_sequence.append(partition)
+                else:
+                    partitioned_sequence.append(partition[0])
+                    partitioned_sequence.append(partition[1:])
+
```
gwwatkin, line 17: You didn't like [the concise implementation based on `.split("H")`](https://replit.com/@gwwatkin/partitiongatesequence)? I thought it would have been cleaner, there is quite a bit of logic here  :sweat_smile: .

Note that you would need to add X handling, but I think you could add it pretty concisely and get the 'HXH' case which might be nice to have since the gate sequences keep changing

tests/gates/compress_rotation_approximations_test.py
```diff
@@ -11,8 +11,11 @@ class TestGateCompression:
         [
             ("SSSXHSTHTHSTHTH", ["SSS", "X", "HSTH", "T", "HSTH", "T", "H"]),
             ("SHSSXH", ["S", "H", "SS", "X", "H"]),
+            ("SHSSHX", ["S", "HSSH", "X"]),
             ("HSSTTSHSX", ["HSSTTSH", "S", "X"]),
             ("STXH", ["ST", "X", "H"]),
+            ("HSTX", ["H", "ST", "X"]),
+            ("XHST", ["X", "H", "ST"]),
```
gwwatkin, line 18: We already have a list case, don't you want to add them all?
```
print(partition_gate_sequence('H'))
print(partition_gate_sequence('S'))
print(partition_gate_sequence('HS'))
print(partition_gate_sequence('SH'))
print(partition_gate_sequence('HSH'))
print(partition_gate_sequence('SHS'))
print(partition_gate_sequence('HSHS'))
print(partition_gate_sequence('SHSH'))
print(partition_gate_sequence('HSHSH'))
print(partition_gate_sequence('SHSHS'))
print(partition_gate_sequence('HSHSHS'))
print(partition_gate_sequence('SHSHSH'))
```

tests/gates/snapshots/approximate_test/test_approximate_rz_with_compression/phase1/gates.txt
```diff
@@ -1 +1 @@
-[]
\ No newline at end of file
+[S(target_qubit=1)]
```
gwwatkin: Shouldn't this be `[PauliRotations(target_qubit=1, phase=Fraction(1, 2), axis=Z)]`? maybe we do have a double call to `to_clifford_plus_t`

src/lsqecc/gates/compress_rotation_approximations.py
gwwatkin: Ok I see some of the subtlety now, you would also have to modify the split solution based to get the behavoiur you described in your comment. I think you can decide best which to go with :slightly_smiling_face:

gwwatkin (commented):
> Right now, partition_gate_sequence breaks down everything that is not on the same basis and from_gate_string assumes the strings it receives are already passed through the partition_gate_sequence function! 

Makes sense, reasonable invariant.

> I pushed some changes now with some bug fixes, but below are some examples that explain the expected behavior:
"SSSXHSTHTHSTHTH" --> ["SSS", "X", "HSTH", "T", "HSTH", "T", "H"]
("STXH", ["ST", "X", "H"]),
("HSTX", ["H", "ST", "X"]),
("XHST", ["X", "H", "ST"]),

`"XHST", ["X", "H", "ST"]` Yes, maybe this makes more sense than the implementation I suggested, which would group  HST. :thinking: 

> If the behavior is too ambiguous, let's make any design changes necessary that are necessary

Tricky issue, I don't think it's ambiguous but subtle. I made a comment asking about why you went with this hard coded implementation over the one I suggested, but now I see it might be because of these corner cases. As long as you have well define behavior for all the examples I posted I think whichever implementation is good.

src/lsqecc/gates/compress_rotation_approximations.py
isolatedinformation: The concise implementation was nice :) I read the comments from bottom-up, last time! so by the time I got to it, I already fixed some bugs and was just too lazy to change it!

tests/gates/compress_rotation_approximations_test.py
isolatedinformation: Yeah I will do that

tests/gates/snapshots/approximate_test/test_approximate_rz_with_compression/phase1/gates.txt
isolatedinformation: It's not a double call! The logic in `approximate.py`: calls `from_gate_string` only if the length of the partition is larger than 1. If it is just one, it uses the same code as the `compress_rotations = False` case. Giving out H, S and T gates respectively

src/lsqecc/gates/compress_rotation_approximations.py
isolatedinformation: I tried modifying the split behavior this morning, but it would require gate-by-gate processing like the code already exists, and becomes as complex as the code that exists to my knowledge and the edge cases are not handled properly or rather I am not able to come up with a cleaner solution! 
If we want to modify cases like `HSSSS` or while encountering an `X`. The main idea behind my code, I keep adding to a partition/subsequence as long as I am in the same basis. Once I encounter a change of basis, either an X or a second H, I pop out the subsequence/partition to the list of partitions.

tests/gates/snapshots/approximate_test/test_approximate_rz_with_compression/phase1/gates.txt
isolatedinformation: Made changes in the latest push and modified the behavior! The logic was that it was passing to the `from_gate_string` only if the string was longer than length 1! Another question, should the single Hadamards also be converted to a rotation! Right now, I leave them as it is.

gwwatkin: Oh I see, if it was intentional there is no problem. Hadamards can stay hadamards because we can do them transversally :+1:

gwwatkin (commented):
Tough one. I think we are pretty much good now.

 One last think to consider is the half angle conversion for Pauli Rotations. Right now all the angles are double those in the convention of the latest lattice surgery papers and our existing block based Pauli rotations. I think we should also switch here to half angles

isolatedinformation:
@gwwatkin maybe  it would be better to make a new PR for the half-angle change so that this pr does not become too large!

gwwatkin (approved):

PR #303, opened 2022-07-29
Adds docstrings to EdgeType class

dayvidpham:
I'm not exactly sure:

- if the stitched EdgeTypes are the same Pauli operator, and
- what the AncillaJoin enum value means-- if what I wrote is wrong, feel free to let me know and I can rewrite it.

Note: Docstrings had to be added below each enum value to make Sphinx happy. Putting the docstring above is possible if I use `#:`, but I wanted to stay consistent.

### DIFF
--- a/src/lsqecc/patches/patches.py
+++ b/src/lsqecc/patches/patches.py
@@ -49,6 +49,12 @@ class Orientation(Enum):
 
 
 class EdgeType(Enum):
+    """Enum for representing what an Edge looks like and what type of operation it encodes.
+    When representing Pauli operators, an Edge is (solid or dashed) and (stitched or unstitched).
+    Alternatively, it can represent the border of joining ancilla patches.
+    Different Pauli operators are mapped to different EdgeType values: X is dashed, and Z is solid.
+    """
+
     Solid = "Solid"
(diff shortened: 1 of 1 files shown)


src/lsqecc/patches/patches.py
```diff
@@ -49,20 +49,37 @@ def get_graph_edge(edge: Edge) -> Optional[Tuple[Tuple[int, int], Tuple[int, int
 
 
 class EdgeType(Enum):
+    """Enum for representing what an Edge looks like and what type of operation it encodes.
+    When representing Pauli operators, an Edge is (solid or dashed) and (stitched or unstitched).
```
alexnguyenn, line 53: Stiched just means that the stabilizers are active on the boundary

alexnguyenn (approved):

src/lsqecc/patches/patches.py
gwwatkin: Yeah leaving the word stitched un-documented might be confusing

gwwatkin (changes requested):
Can you clarify stitched as boundary stabilizers being active? Thanks :)

PR #306, opened 2022-07-29
Adds the docs page docs/api/patches

dayvidpham:
This PR is purely for creating and updating the `docs/api/patches` page, but the docs page will be empty until the PRs with class docstrings are merged in.

This is separate since updating `docs/sources/api/patches.md` in each PR that adds a class docstring would create a merge conflict.

### DIFF
--- a/docs/source/api/index.md
+++ b/docs/source/api/index.md
@@ -5,6 +5,6 @@ This section is still under active development
 ```
 
 ```{toctree}
-
+patches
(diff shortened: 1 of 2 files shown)


alexnguyenn (approved):

gwwatkin:
Good to have the auto-gened docs for the exposed API for Patches.

@alexnguyenn is the doc build failing because of a version miss-match?

